PR #139, opened 2016-04-03
Add default config path to systemd services.

stoupa-cz:
Config file is installed by default, so I would expect that it will be used by
the systemd service. Add argument for the configuration to the services.

Signed-off-by: Tomas Novotny <novotny@rehivetech.com>


ralight:
The config file isn't installed by default by the project (it is installed as mosquitto.conf.example), but I would expect distros or users to use a config file so I think this change is fine. Thanks for your contribution.

stoupa-cz:
@ralight So this is inconsistency between Makefile and CMake installation rules. I checked it and it is true that Makefile installs conf with .example suffix. But CMake just copies the conf as it is (w/o .example suffix). I can send you a patch for this simple thing if you want.

PR #140, opened 2016-04-15
Improve performance with lots of queued message

PierreF:
In our use case, we could have large number of queued message for one consumer. In such case we had very poor performance.
After few investigation, I found that the issue was [mqtt3_db_message_write](https://github.com/eclipse/mosquitto/blob/d20355c8ac64746876bdfd0c47f9ac6096bee9dc/src/database.c#L810): each time a message could be sent to client, it iterate over whole list of message for this client.
With thousands of messages Mosquitto starts to use a lots of CPU and message consumption is way slower

This PR add counters to avoid iterating over all queued messages and stop as soon
as counters tell us that no more actions are required for now.

With a test client that:
* create a consumer with clean_session=False, subscribe to "topic" and disconnect
* create a producer that send N message to "topic" and disconnection
* re-connect the initial consumer and consume all N messages

I have the following result:

number of messages | message/s without PR | message/s with PR
------------ | ------------- | -------------
2500 | 16700 message/s | 18100 message/s
7500 | 15400 message/s | 18400 message/s
10000 | 13100 message/s | 19200 message/s
25000 | 3800 message/s | 19100 message/s

### DIFF
--- a/lib/mosquitto_internal.h
+++ b/lib/mosquitto_internal.h
@@ -199,6 +199,8 @@ struct mosquitto {
 	struct mosquitto_client_msg *last_msg;
 	int msg_count;
 	int msg_count12;
+	int msg_count_ready_send;
+	int msg_count_queued2_in;
 	struct _mosquitto_acl_user *acl_list;
 	struct _mqtt3_listener *listener;
 	time_t disconnect_t;
--- a/src/context.c
+++ b/src/context.c
@@ -73,6 +73,8 @@ struct mosquitto *mqtt3_context_init(struct mosquitto_db *db, mosq_sock_t sock)
 	context->last_msg = NULL;
 	context->msg_count = 0;
 	context->msg_count12 = 0;
+	context->msg_count_ready_send = 0;
+	context->msg_count_queued2_in = 0;
 #ifdef WITH_TLS
 	context->ssl = NULL;
 #endif
--- a/src/database.c
+++ b/src/database.c
@@ -222,6 +222,16 @@ static void _message_remove(struct mosquitto_db *db, struct mosquitto *context, 
 	if((*msg)->qos > 0){
 		context->msg_count12--;
 	}
+	if((*msg)->state == mosq_ms_queued && (*msg)->direction == mosq_md_in && (*msg)->qos == 2){
+		context->msg_count_queued2_in--;
+	} else if ((*msg)->state == mosq_ms_publish_qos0 ||
+			(*msg)->state == mosq_ms_publish_qos1 ||
+			(*msg)->state == mosq_ms_publish_qos2 ||
+			(*msg)->state == mosq_ms_resend_pubrel ||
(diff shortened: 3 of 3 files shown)


ralight:
Thanks for this. I'm keen on improving this for sure. 

My initial impression is that I'm not sure this is the best way to approach the problem, but that's possibly because I've only had a very brief chance to review it (I've been on holiday and came back to a broken phone line which is only just fixed).

I wondered about maybe keeping separate arrays for the in-flight and queued messages and whether that could help instead. Like I said, I've not really thought about it properly yet though.

Two things are for sure though - please ensure you have signed the Eclipse CLA as described in CONTRIBUTING.md - and please use the "develop" branch as the base for this work.

PierreF:
It seems also more clear to use 2 queues, one for in-flight and one for queued message.

I've few question regarding current code that I don't understand:

* In [db__message_write](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L805) it count only message with mosq_md_in against the max_infight. All other place seem to count both in and out message. Is it wanted ?
* Also in [same place](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L892), it only dequeue incoming message with QoS==2, not all other kind. Is there a special reason or could it dequeue any message ?

Without this, I'm not sure it could use only 2 queues (in-fight and queued) and it may need 3 queues (in-flight, queued output and queued input).

Also while looking at code, I've found few strange thing:

* At [some place](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L892) there is a check for max_inflight == 0 but [other](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L728) don't. Shouldn't it always check for max_inflight == 0 ?
* In [db__message_delete](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L293), a incoming queued message with QoS==2 is moved to mosq_ms_wait_for_pubrel. From my understanding, it should either send a pubrec before (like [here](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L744)) or changed to state mosq_ms_send_pubrec (like [here](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L894))
* In [db__message_remove](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L243), the counter i is never incremented.

If those points are bugs, I could fix them while splitting the messages queue.

PierreF:
Created new PR #141 based on develop branch. Did it using two arrays instead of counters.

PR #141, opened 2016-04-18
Improve performance with lots of queued message

PierreF:
Split message queue in two queues: in-flight and queued to avoid the
need to iterate over all messages.

See #140 for complete description. It's the same PR but:

* based on develop instead of master
* using 2 chained-list instead of counter
* with CLA signed :)

About my question on current code, I've assumed that:

* [db__mesage_write](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L892) could dequeue all kind of message (not only incoming message) and that max_inflight is for both incoming and outgoing message
* max_inflight should be always check for zero value (unlimited)
* In [db__message_delete](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L293), send a pubrec like in db__message_release and db__message_write
* The dequeue in [db__message_remove](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/database.c#L243) is unnecessary, as all called do a dequeue anyway.

Also I've spotted a possible bug: [handle_connect](https://github.com/eclipse/mosquitto/blob/62402f7b60c17d9dc074b0ccd53a7a32c906994e/src/handle_connect.c#L546) : when checking for possible ACL change, it check to update ```msgs``` pointer, but don't for ```last_msg``` pointer.

### DIFF
--- a/lib/mosquitto_internal.h
+++ b/lib/mosquitto_internal.h
@@ -195,8 +195,10 @@ struct mosquitto {
 	bool is_dropping;
 	bool is_bridge;
 	struct mosquitto__bridge *bridge;
-	struct mosquitto_client_msg *msgs;
-	struct mosquitto_client_msg *last_msg;
+	struct mosquitto_client_msg *inflight_msgs;
+	struct mosquitto_client_msg *last_inflight_msg;
+	struct mosquitto_client_msg *queued_msgs;
+	struct mosquitto_client_msg *last_queued_msg;
 	int msg_count;
 	int msg_count12;
 	struct mosquitto__acl_user *acl_list;
--- a/src/context.c
+++ b/src/context.c
@@ -70,8 +70,10 @@ struct mosquitto *context__init(struct mosquitto_db *db, mosq_sock_t sock)
 		}
 	}
 	context->bridge = NULL;
-	context->msgs = NULL;
-	context->last_msg = NULL;
+	context->inflight_msgs = NULL;
+	context->last_inflight_msg = NULL;
+	context->queued_msgs = NULL;
+	context->last_queued_msg = NULL;
 	context->msg_count = 0;
 	context->msg_count12 = 0;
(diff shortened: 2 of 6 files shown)


ralight:
I need a bit of time to look at this properly to understand what you're doing exactly. I'm afraid I've got really urgent things at work for the next 10 days so it's likely I won't be doing anything in the evenings. Please feel free to nudge me after then.

The handle_connect lack of `last_msg` check does look like a bug to me as well.

To answer other query you had, the inflight message handling should be done one of two ways - either the outgoing messages only, or both incoming and outgoing messages separately. I think this needs a bit wider discussion.

Something I've looked at in the past is completely separating the incoming and outgoing message queues. I can't remember what stopped me implementing it, but would be keen to do that as well.

markaswift:
Nudge ;)

PierreF:
Any update? Did you had time to look at this PR ?

ralight:
So far I've found that the code fails on one of the bridge tests, but I've not figured out why yet.

Try `make test` and see whether it works for you.

PierreF:
make test works for me, I have no error.
```
$ git clean -fxd
$ make test
[...]
./05-clean-session-qos1.py 
./06-bridge-reconnect-local-out.py
./06-bridge-br2b-disconnect-qos1.py
./06-bridge-br2b-disconnect-qos2.py
./06-bridge-b2br-disconnect-qos1.py
./06-bridge-b2br-disconnect-qos2.py
./07-will-qos0.py
[...]
```

ralight:
Ok, I can't see what was causing this problem it must have been unrelated as I can't reproduce it now. This looks fine and I've had it running on test.mosquitto.org for a while as well now.

PR #143, opened 2016-04-19
CMake changes : static library target, no /sbin/ldconfig on Mac, and Documentation option

anonymouse64:
This pull request has a couple of changes to the CMake build system.

* Now has an additional target `libmosquitto_static` which compiles the same files as libmosquitto into a static archive
* When building the new `libmosquitto_static` the declaration define `libmosq_EXPORT` also gets null defined as all of the functions are contained in the static library
* No longer runs `/sbin/ldconfig` on Mac OS X (specifically is now run when `UNIX` is true, and `APPLE` is false)
* Adding a `DOCUMENTATION` option that enables/disables copying the man files during `install` target

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -20,8 +20,7 @@ include_directories(${mosquitto_SOURCE_DIR} ${mosquitto_SOURCE_DIR}/lib
 			${OPENSSL_INCLUDE_DIR} ${PTHREAD_INCLUDE_DIR})
 link_directories(${mosquitto_SOURCE_DIR}/lib)
 
-add_library(libmosquitto SHARED
-	logging_mosq.c logging_mosq.h
+set(C_SRC logging_mosq.c logging_mosq.h
 	memory_mosq.c memory_mosq.h
 	messages_mosq.c messages_mosq.h
 	mosquitto.c mosquitto.h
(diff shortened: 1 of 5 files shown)


anonymouse64:
After opening this pull request I signed the CLA via eclipse.org, I'm assuming the system will pick it up or do I need to reopen the pull request to trigger the ip-validation?

ralight:
I'm in favour of these changes, but would appreciate some tweaks. I'll comment on the specifics in the commits but my general requests are please base this work on the develop branch, and please use indentation.

I would have checked your CLA status even if just accepting as is, but it should get picked up again next time.

lib/CMakeLists.txt
```diff
@@ -41,6 +41,28 @@ add_library(libmosquitto SHARED
 	util_mosq.c util_mosq.h
 	will_mosq.c will_mosq.h)
 
+#target for building static version of library
```
ralight: I think it would be best here to define say

set(SRC logging_mosq.c logging_mosq.h ...)

add_library(libmosquitto SHARED ${SRC})
add_library(libmosquitto_static STATIC ${SRC})

src/CMakeLists.txt
```diff
@@ -109,13 +109,11 @@ endif(NOT APPLE)
 
 target_link_libraries(mosquitto ${MOSQ_LIBS})
 
-if (UNIX)
```
ralight: I believe this simplification is incorrect because it will break the build on Windows.

anonymouse64: Ah, I forgot about that. I will revert this change

lib/CMakeLists.txt
anonymouse64: I agree, that would reduce the duplication here. I elected to name it `C_SRC` however because here it's just all the C code, and so if one wanted to expand the C++ to use this convention, one could use a similar name such as `CXX_SRC` or something.

anonymouse64:
Closing this pull request and reopening [here](https://github.com/eclipse/mosquitto/pull/144) on develop, as I'm unsure how to change the destination branch on GitHub.

PR #144, opened 2016-04-19
CMake improvements

anonymouse64:
* Now has a `DOCUMENTATION` option for enabling/disabling documentation man pages
* Created a new static library target for libmosquitto, named `libmosquitto_static`
* Remove all function export declarations for the static library
* No longer calls `/sbin/ldconfig` on platforms other than Unix

(this is basically the same as [this PR](https://github.com/eclipse/mosquitto/pull/143))

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,8 @@ endif (${WITH_SOCKS} STREQUAL ON)
 
 option(WITH_SRV "Include SRV lookup support?" OFF)
 
+option(DOCUMENTATION "Build documentation?" ON)
+
 # ========================================
 # Include projects
 # ========================================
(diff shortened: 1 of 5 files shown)


ralight:
This looks good, I've just one more thing I'm not sure about and can't check right now. What happens with the mosq_EXPORT dll define when the libraries are installed and are outside of cmake? Do we have to assume that the dev using it will define STATIC_LIB?

anonymouse64:
That's a good point, I hadn't quite thought of that. I suppose there are two ways we could handle it, either make mosquitto.h be processed and configured by CMake to have the defines show up as appropriate in the end file, or else we could document the option so people could be expected to define `STATIC_LIB`. Not sure which you would prefer.

ralight:
Thinking about it, documenting the need for `STATIC_LIB` seems like the best choice. This particular issue only affects the people using Windows that also want to use a static library.

It would probably be sensible to make it a bit more project specific, maybe LIBMOSQUITTO_STATIC.

anonymouse64:
Okay, so I imagine that at the very least these two files should be updated:

https://github.com/eclipse/mosquitto/blob/master/compiling.txt

https://github.com/eclipse/mosquitto/blob/master/readme-windows.txt

Are there other places in the repository that should be updated with this change? Maybe a comment in lib/CMakeLists.txt?

Issue #18, opened 2016-03-15
CONNACK Message not sent for Unauthorized Connect Request for Websocket


ralight:
migrated from Bugzilla [#484761](https://bugs.eclipse.org/bugs/show_bug.cgi?id=484761)
status UNCONFIRMED severity _critical_ in component _Mosquitto_ for _1.4_
Reported in version _1.4_ on platform _PC_
Assigned to: Roger Light

On 2015-12-21 08:43:30 -0500, koray sariteke wrote:

> For Authentication, we implemented mosquitto_plugin.h functions.
> 
> When we Start mqtt broker in Websocket protocol, CONNACK with result code 5 IS NOT sent back to client. TCP client is disconnected by Mosquitto Server without information.
> 
> Up to MQTT RFC
> http://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html#_Toc398718035
> 
> If a well formed CONNECT Packet is received by the Server, but the Server is unable to process it for some reason, then the Server SHOULD attempt to send a CONNACK packet containing the appropriate non-zero Connect return code from this table. If a server sends a CONNACK packet containing a non-zero return code it MUST then close the Network Connection [MQTT-3.2.2-5].

On 2015-12-21 09:02:10 -0500, koray sariteke wrote:

> vi read_handle_server.c +364
> # ifdef REAL_WITH_TLS_PSK
> 
> ```
>             }
> ```
> # endif /\* REAL_WITH_TLS_PSK */
> 
> ```
>     }else{
> ```
> # endif /\* WITH_TLS */
> 
> ```
>             if(username_flag){
>                     rc = mosquitto_unpwd_check(db, username, password);
>                     switch(rc){
>                             case MOSQ_ERR_SUCCESS:
>                                     break;
>                             case MOSQ_ERR_AUTH:
>                                     _mosquitto_send_connack(context, 0, CONNACK_REFUSED_NOT_AUTHORIZED);
>                                     mqtt3_context_disconnect(db, context);
>                                     rc = 0;
>                                     goto handle_connect_error;
>                                     break;
> ```
> 
> if mqtt3_context_disconnect(db, context); line removed and rc set to 0, CONNACK sent to client with reason code 5 but TCP connection stays open which is not true state.

ramarro123:
@ralight i know that this bug it's starting to age as a good wine, but i am pinging you to know if it will be closed as a non issue or if it's still in work.

[...]

PR #146, opened 2016-04-23
Fixes for bug [484761]

focusware-develop:
Previous pull request broke the build (some tests failing, because of a different return value when the authentication failed).

### DIFF
--- a/lib/mosquitto_internal.h
+++ b/lib/mosquitto_internal.h
@@ -130,6 +130,7 @@ struct _mosquitto_packet{
 	uint16_t mid;
 	uint8_t command;
 	int8_t remaining_count;
+	bool force_synchronous;
 };
 
 struct mosquitto_message_all{
--- a/lib/net_mosq.c
+++ b/lib/net_mosq.c
@@ -154,19 +154,40 @@ int _mosquitto_packet_queue(struct mosquitto *mosq, struct _mosquitto_packet *pa
 	packet->pos = 0;
 	packet->to_process = packet->packet_length;
 
+	bool skip_queue = false;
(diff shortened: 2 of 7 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=484761

ralight:
The github issue for this bug is https://github.com/eclipse/mosquitto/issues/18

I'm afraid the fix you're proposing is invalid. If it was this easy it would have been fixed already. libwebsockets forbids calling libwebsocket_write/lws_write except from within the defined callback. It might work for you, but it's not the right thing to do.

The change to use CONNACK_REFUSED_BAD_USERNAME_PASSWORD is also incorrect, the spec defines this as for the situation when "the data in the user name or password is malformed".

focusware-develop:
Ok, I understand your point. The problem with the current implementation is that the (SSL) socket connection is closed before the CONNACK is sent over the connection, and therefore it is not arriving at the client side.

Is there maybe a way to mark the connection as "to-be-closed a little later", instead of calling mqtt3_context_disconnect() directly? Or have the context disconnect function flush the buffers before closing the connection?

focusware-develop:
About the CONNACK_REFUSED_BAD_USERNAME_PASSWORD change, you are absolutely right :-)
I only read the return code response, but the description indeed mentions "the data in the user name or password is malformed". Sorry about that.

ralight:
Yes absolutely there is a way to do it - but I'm not entirely sure how. I think the route to the solution is to add an extra state in for the client which is "closing, but only after we've sent outgoing data".

Issue #154, opened 2016-05-12
TLS bridge not re-connecting on Windows


dkaukov:
mosquitto version 1.4.8. Bridge is successfully connected on start, but once connection is dropped it is not able to re-connect again. To reconnect I have to re-start server. It is working fine for non-TLS bridge connection and also works fine in Linux build in same environment

Log:
```
1463029685: mosquitto version 1.4.8 (build date 14/02/2016 15:33:31.09) starting
1463029685: Config loaded from mosquitto.conf.
1463029685: Opening ipv6 listen socket on port 1883.
1463029685: Opening ipv4 listen socket on port 1883.
1463029685: Bridge local.Prod-213213123213- doing local SUBSCRIBE on topic XXXXXXXXXXXXX/+/PUB/#
1463029685: Bridge local.Prod-213213123213- doing local SUBSCRIBE on topic XXXXXXXXXXXXX/+/REP/#
1463029685: Bridge local.Prod-213213123213- doing local SUBSCRIBE on topic XXXXXXXXXXXXX-PP/+/REP/#
1463029685: Connecting bridge ot (xxx.com.au:8883)
1463029685: Bridge Prod-213213123213- sending CONNECT
1463029686: Received CONNACK on connection local.Prod-213213123213-.
1463029686: Bridge local.Prod-213213123213- sending UNSUBSCRIBE (Mid: 43, Topic: XXXXXXXXXXXXX-PP/+/PUB/#)
1463029686: Bridge local.Prod-213213123213- sending UNSUBSCRIBE (Mid: 44, Topic: XXXXXXXXXXXXX-PP/+/REP/#)
1463029686: Bridge local.Prod-213213123213- sending UNSUBSCRIBE (Mid: 45, Topic: XXXXXXXXXXXXX-PP/+/REP/#)
1463029686: Received UNSUBACK from local.Prod-213213123213-
1463029686: Received UNSUBACK from local.Prod-213213123213-
1463029686: Received UNSUBACK from local.Prod-213213123213-
1463029699: Socket error on client local.Prod-213213123213-, disconnecting.
1463029704: Bridge local.Prod-213213123213- doing local SUBSCRIBE on topic XXXXXXXXXXXXX/+/PUB/#
1463029704: Bridge local.Prod-213213123213- doing local SUBSCRIBE on topic XXXXXXXXXXXXX/+/REP/#
[...]

PR #160, opened 2016-05-13
Workaround for Windows SSL

dkaukov:
In windows OpenSSL sometimes returns SSL_ERROR_SYSCALL on non-blocked socket when connection in progress. This is workaround for this situation.


ralight:
Does this fix the problems with bridges not reconnecting that you reported?

dkaukov:
Yes, it is regarding https://github.com/eclipse/mosquitto/issues/154

PR #161, opened 2016-05-15
Sending  signals to mosqitto on NT, using named events.

dkaukov:
Similar to the:
https://github.com/apache/httpd/blob/8322599c746bbdf1410a098a5d4764499baf7670/server/mpm/winnt/mpm_winnt.c#L156

Signed-off-by: Dmitry Kaukov <dkaukov@opentable.com>

### DIFF
--- a/src/mosquitto.c
+++ b/src/mosquitto.c
@@ -184,6 +184,59 @@ void handle_sigusr2(int signal)
 	flag_tree_print = true;
 }
 
+/*
+* Signalling mosquitto on NT.
+*
+* Under Unix, mosquitto can be told to shutdown or restart by sending various
+* signals (HUP, USR, TERM). On NT we don't have easy access to signals, so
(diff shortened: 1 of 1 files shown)


ralight:
I'm not sure that I can accept this. If you read the "Due Diligence Procedures" at [1](https://www.eclipse.org/legal/committerguidelines.php) then you'll see that by adding the Signed-off-by line you are indicating this change is in compliance with the [Contributor’s Certificate of Origin](https://www.eclipse.org/legal/CoO.php), the first point of which is "I have authored 100% of this contribution."

It is clear that at least some of this change has been copied from the link that you provided, unless you authored the code at Apache which it doesn't appear is the case.

I'm sorry, I'm not trying to be awkward but we have to take IP requirements seriously.

dkaukov:
Hi, I just used same naming convention for events. Is it also IP?

PR #163, opened 2016-05-16
Signalling mosquitto process on Win32, using named events:

dkaukov:
mosqPID_shutdown
mosqPID_reload
mosqPID_backup
mosqPID_vacuum

where PID is the PID of the mosquitto process


dkaukov:
Hi, I have changed:

- Event prefix to "mosq"
- Comments

ralight:
Thanks and sorry for the delay. I'll be changing things ever so slightly - there is no need for vacuum now for example.

Issue #170, opened 2016-05-25
bind address for websockets


minghuadev:
Hi, Trying to run mosqitto on openshift which would allow only to use a virtual IP address thus need to bind to an IP address in addition to a port. It failed. Is there a way of doing it? 

Googled an email by Chris on Sep2 2015 on the dev mainling list suggesting use of "iface" parameter. Has that been done? 

Following the information from the mailing list post eventually got this diff to have made it work: 

    diff --git a/src/websockets.c b/src/websockets.c
    index d79fabf..839f314 100644
    --- a/src/websockets.c
    +++ b/src/websockets.c
    @@ -627,6 +627,18 @@ struct libwebsocket_context *mosq_websockets_init(struct _mqtt3_listener *listen
     
            lws_set_log_level(log_level, log_wrap);
     
    +    if ( listener->host ) {
    +        size_t sz = strnlen(listener->host, 16);
    +        if ( sz >=7 && sz <= 16 ) { /* IP in string: 7 min length. 15 max. */
    +            info.iface = listener->host;
    +               _mosquitto_log_printf(NULL, MOSQ_LOG_INFO, 
    +                        "Listener host \"%s\" of size %zu.", listener->host, sz);
    +        } else {
    +               _mosquitto_log_printf(NULL, MOSQ_LOG_WARNING, 
    +                        "Warning: Malformed listener host \"%s\" of size %zu.", 
    +                                  listener->host, sz);
    +        }
    +    }
            _mosquitto_log_printf(NULL, MOSQ_LOG_INFO, "Opening websockets listen socket on port %d.", listener
            return libwebsocket_create_context(&info);
     }

Does this or something similar need to be patched to mosquitto? 

Thanks.
-Minghua

ralight:
I think a better approach would to just add `info.iface = listener->host` and have listeners for websockets just pass an interface name instead of an ip address/hostname.

minghuadev:
Hi, Roger, Yes making it simply `info.iface = listener->host` looks better but I don't have the knowledge right now to assume that's safe enough. 

[...]

PR #171, opened 2016-05-26
allow websocket bind to ip address. issue #170

minghuadev:
the change based on discussion on issue #170


minghuadev:
close as it's to a wrong branch. and the change is simple enough not to need a pull request.

PR #179, opened 2016-05-31
Enhance static library build

Lance0312:
1. Fix a build fail which is caused by a lack of source object in libmosquitto.
2. Add an CMake option BUILD_STATIC_LIBRARY to allow control over building static libraries or not.
3. Add an CMake option WITH_PIC to allow control over building static libraries with PIC or not.
4. Add build configuration to build CPP static library.


Lance0312:
I was wondering if you guys prefer
1) the CPP static library a superset of the C static library, or
2) the CPP static library only contains object file of mosquittopp.cpp, and requires users to link both the CPP static library and the C static library to their programs?

This pull request goes with 1)

ralight:
I don't have any strong feelings on the CPP library, let's see if anybody else chimes in.

Thanks, this generally looks good, I'll give it a test later.

Lance0312:
Thank you!

PR #181, opened 2016-06-01
Rewrite Makefiles under test/ to CMake script

Lance0312:
These are the Makefiles located under test/
```
./test/Makefile
./test/lib/Makefile
./test/lib/c/Makefile
./test/lib/cpp/Makefile
./test/broker/Makefile
./test/broker/c/Makefile
```
This pull request rewrite the above Makefiles to CMakeLists.txt files which provide identical functions at best.
```
./test/CMakeLists.txt
./test/lib/CMakeLists.txt
./test/lib/c/CMakeLists.txt
./test/lib/cpp/CMakeLists.txt
./test/broker/CMakeLists.txt
./test/broker/c/CMakeLists.txt
```


ralight:
I don't wish to dampen your enthusiasm, but the bare Makefiles are the supported build tool for Linux. The CMake support is only there for Mac and Windows support. I like what you've done, but it's increasing the maintenance burden. It's already possible to run tests through cmake with "make Tests".

My question then is what value does this commit bring over the existing support?

Lance0312:
It confuses me at the first glance that both bare Makefiles and CMake build scripts exist. I thought this project is in the progress of migrating from bare Makefiles to CMake. The travis-configure.sh file is the only thing which gives a hint that those CMakeLists.txt files are only for Mac OS. If maintenance effort is what are concerned with, I believe that using just one build technique is the better way to go.

Also, having both bare Makefiles and CMake build scripts means doubling the work for contributors and inconsistent features among platforms. One example is eclipse#179. Static library building is supported long ago in bare Makefiles, not until now it is made into CMakeLists.txt. Although it could mean no Mac/Windows user has the need of it (ps. I am actually a linux user), I expect a more positive attitude on dealing with this sort of things.

 Thus, I would like to propose a fully migration to CMake.

ralight:
I've not got a negative attitude, I was simply asking what you thought the benefit was - when we had different views of the situation. You can see from my point of view there is little benefit having cmake based tests as well as the Makefile tests, just as I now understand the way you were thinking.

config.mk explains what the Makefiles are for. compiling.txt and CMakeLists.txt explain that cmake is only intended for Windows and Mac.

I agree that having a single build script would be best. The trouble is that if there was only a single build script it would be the bare Makefiles. This is what I use all of the time for development. The cmake files exist only to provide support for Windows and Mac.

Lance0312:
I was not saying that you got a negative attitude, apologize for my wording. What I would like to express is that as long as bare Makefiles and CMakeLists.txt files both exist, inconsistency of build features will eventually appear, unless either the contributors or the maintainers put their efforts on solving it. Since CMake is used for supporting building on Mac/Windows, and there are already many big projects out there using CMake, I would think that it would be nice to migrate to CMake. If it is your final decision to stick to bare Makefiles, I suggest merging compiling.txt into readme.md or giving a reference to compiling.txt in readme.md. And inform contributors to make their contribution apply to both bare Makefiles and CMakeLists.txt as far as possible.

BTW, thanks for the heads up on compiling.txt, totally missed that file :O

ralight:
I agree that having the two scripts means there can be inconsistencies, and that is unfortunate but just using cmake would make life more difficult for me on a day to day basis.

I accept that merging the information into the readme.md and improving it would be a good idea, it is possible to miss right now.

Lance0312:
Thank you. I will close eclipse#181 and eclipse#182.

PR #182, opened 2016-06-01
Add Travis CI jobs to build and test with CMake in linux

Lance0312:
This pull request adds jobs to the Travis CI build matrix to run builds and tests with CMake in linux. These jobs define a environment variable LINUX_CMAKE. And travis-configure.sh configures the build with CMake if the environment variable is set.

Note: The build environment is changed to Ubuntu Trusty, since Ubuntu Precise ships with CMake 2.8.7, and the current build script requires at least CMake 2.8.11 to work.


Lance0312:
Please note that this pull request depends on eclipse#181, so there are some irrelevant commits shown.

PR #183, opened 2016-06-02
Enhance bare makefile static library build

Lance0312:
This pull request includes two changes
1. Install the C static library
2. Build and install the CPP static library

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -80,6 +80,9 @@ WITH_DOCS:=yes
 # Build with client support for SOCK5 proxy.
 WITH_SOCKS:=yes
 
+# Build static libraries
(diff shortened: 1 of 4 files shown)


chelliwell:
Wouldn't it be better to have an additional/optional install target for the static library...? (I'm cross-compiling/installing, and don't want or need the static lib on my target)

Lance0312:
Of course, I'll add the switch later today.

ralight:
Thanks for this, I've now merged your work but have changed the behaviour a bit. I feel as though the best approach is actually to keep the two static libraries separate.

Lance0312:
That works for me as well, thank you!

PR #196, opened 2016-06-27
[189] Adds fflush before fsync call to help guarantee database validity on disk even in event of power failure.

kcallin:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=189

karlp:
what's the reasoning for dropping the fsync on the directory?

kcallin:
I understand that the goal of this code is to guarantee that mosquitto.db always contain valid data.  The fsync call on the directory will ensure the directory entry is on disk before the close operation but this is not a guarantee that is required for that goal.  If power is lost before the close but without the directory fsync then the mosquitto.db.new entry may not exist.  In this case mosquitto will continue to use the old "mosquitto.db" file (which is okay).  The rename operation occurring a few lines later will operate on what is in cache so the directory fsync is not useful there either.

If the rename operation is interrupted the file system will contain one of:

Possiblity 1 (OK):
mosquitto.db (i-node X)
mosquitto.db.new (i-node Y)

Possibility 2 (Not OK):
mosquitto.db (i-node Y)
mosquitto.db.new (i-node Y)

Possiblity 3 (OK):
mosquitto.db (i-node Y)

You know, as I typed this out I discovered that possibility 2 was not okay!  We should add a remove call before opening mosquitto.db.new to ensure that a valid mosquitto.db is not truncated on open.

kcallin:
I added the required unlink call.  It needs to sit and test for some time to mechanically verify that the problem is fixed.

In the meantime, does the patch look reasonable?

ralight:
Any news on the testing? The patch looks ok to me.

ralight:
You'll have to sign the CLA though, and ensure that the commits have a valid signed-off-by entry, as is described in `CONTRIBUTING.md`.

kcallin:
Testing so far does not show any regressions.  To show demonstrate actual improvement I'm looking for 100k successful cycles on a power-cycle rig.  It will take ~2 weeks to finish.

Since no regressions are showing up so far it might be best to merge the patch before testing is complete in order to keep it fresh.  If anything shows up after two weeks of testing I can introduce a follow-on patch.

I will read CONTRIBUTING.md and follow the instructions outlined there.

kcallin:
Closing pull request and submitting new pull request that complies with CLA as described in CONTRIBUTING.md.

PR #198, opened 2016-06-29
Fixes a bug where C++ wrapper would always claim SOCKS was not supported

jbwdevries:
The WITH_* flags are not handed over to the C++ wrapper, instead it relies on
the actual library to check status.

Signed-off-by: Johan de Vries (Ubuntu VM) <devries@wivion.nl>


ralight:
Thanks very much!

PR #200, opened 2016-07-01
Implement byte limits for message queues

karlp:
First round draft for discussion:  Some of these pieces are independent, such as the test remaining_length fix, and the db_dump tweaks and can clearly be put into separate pull requests.  Three mains changes:
* broker stats on bytes stored, not just counts
* max_queued_bytes setting now (substantially easier to limit the memory consumed by each client)
* queue_qos0_messages fixed to not infinitely queue.

Questions:  

* Should max_inflight_bytes be documented out and user tunable?  Probably for consistency?
* I had to publish initial load average stats to make it sane for parsing the $SYS tree for testing.  It's still very hard to automatically test some of this, without access to the client level queue stats from the tests.  The broker bytes store is influenced not just by messages to clients, but by the specific sizes of $SYS messages as well!
* likewise, to make a test for sending multiple messages to offline subscribers, using the raw packet interface in mosq_test proved extremely unwieldy.  However, my new test adds a dependency on the paho.mqtt client packages as a downside.

### DIFF
--- a/src/database.c
+++ b/src/database.c
@@ -125,6 +125,7 @@ void db__msg_store_remove(struct mosquitto_db *db, struct mosquitto_msg_store *s
 		}
 	}
 	db->msg_store_count--;
+	db->msg_store_bytes -= store->payloadlen;
 
 	mosquitto__free(store->source_id);
 	if(store->dest_ids){
@@ -556,6 +557,7 @@ int db__message_store(struct mosquitto_db *db, const char *source, uint16_t sour
 	temp->dest_ids = NULL;
 	temp->dest_id_count = 0;
 	db->msg_store_count++;
+	db->msg_store_bytes += payloadlen;
 	(*stored) = temp;
 
 	if(!store_id){
--- a/src/database.c
+++ b/src/database.c
@@ -26,7 +26,68 @@ Contributors:
 #include "time_mosq.h"
 
 static int max_inflight = 20;
+static unsigned long max_inflight_bytes = 0;
 static int max_queued = 100;
+static unsigned long max_queued_bytes = 0;
+
+/**
+ * Is this context ready to take more in flight messages right now?
+ * @param context the client context of interest
+ * @param qos qos for the packet of interest
+ * @return true if more in flight are allowed.
+ */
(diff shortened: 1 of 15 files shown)


karlp:
Note to self: missing changelog entry for queue_qos0_messages fix

src/database.c
```diff
@@ -384,7 +459,10 @@ int db__message_insert(struct mosquitto_db *db, struct mosquitto *context, uint1
 			return 2;
 		}
 	}else{
-		if(max_queued > 0 && context->msg_count12 >= max_queued){
+		if (db__ready_for_queue(context, qos)){
+			log__printf(NULL, MOSQ_LOG_DEBUG, "  off: queueing");
```
ralight: Is this a debugging remnant?

ralight:
First glance (without checking it out) this looks pretty reasonable.

For the initial load messages - a simpler change would be to set `static time_t last_update = 1;` and `static double msgs_received_load1 = -1;` (along with all the rest).

src/database.c
karlp: yep, will kill.

karlp:
for the initial load messages, using last_update = 1 and all the static xxx_load1,5,15 to -1 works, tests all work, but the socket and connection figures are initially -0.00?

```
$SYS/broker/load/bytes/sent/1min 0.00
$SYS/broker/load/bytes/sent/5min 0.00
$SYS/broker/load/bytes/sent/15min 0.00
$SYS/broker/load/sockets/1min 0.00
$SYS/broker/load/sockets/5min -0.00   <<<<<<<<<<<<<<
$SYS/broker/load/sockets/15min -0.00 <<<<<<<<<<<<<
$SYS/broker/load/connections/1min 0.00
$SYS/broker/load/connections/5min -0.00 <<<<<<<<<<<<
$SYS/broker/load/connections/15min -0.00 <<<<<<<<<<<<<<
```

Any other ideas?

karlp:
How about this method of forcing the initial publish of loads?

karlp:
build checks are missing that this kills the build with websockets enabled.  will fix

karlp:
This has been rebased on current develop branch to catch up with the broker_internal.h changes.

Issue #189, opened 2016-06-20
Corruption of persistent database file cause by sudden lost of power


thanhvtruong:
It looks like when the persistent file is being save there is a tiny amount of time where a sudden power lost will cause the persistent database file to be corrupted. 

Thousands of sudden power lost on our system were performed we notice the following:  
- Mosquitto start up with "invalid argument" and "database" read error.
- mosquitto.db and mosquitto.db.new both existed in /var/lib/mosquitto
- Both mosquitto.db and mosquitto.db.new have the same inode. (suggest that a rename has occurred)

We have a similar problem with another application that we build and reading into Linux documentation, we found out that flushing or closing a file is not enough to write the content to disk.  An fsync need to be perform to confirm that the content is written to disk.

You can see the documentation in the man page (man close, under "note" second paragraph) on Fedora 23.

"A successful close does not guarantee that the data has been successfully saved to disk,  as  the  kernel defers writes.  It is not common for a filesystem to flush the buffers when the stream is closed.  If you need to be sure that the data is physically stored, use fsync(2).  (It will depend on the  disk  hardware at this point.)"

ralight:
Thanks for the report, I've added code to call fsync() on the file and on its directory. Could you please confirm whether this fixes the problem for you?

kcallin:
Recommend calling fflush before fsync to ensure that application buffers are completely flushed to kernel buffer before being flushed to disk.  It's awful hard to reproduce this, but between thanhvtruong and myself we started a long-term test series to mechanically verify.

I do not believe the directory sync is required; the rename logic should work as-is.

I opened a pull requrest for these changes and will update as the long-term tests progress.

ralight:
Thanks very much for your work on this, I'm closing this now based on your pull request.

PR #206, opened 2016-07-06
[189] Mosquitto database corrupted on power-loss.

kcallin:
Mosquitto database writes are not atomic and if power is lost during
a write the file will be permanently lost.  This commit makes writes as
atomic as possible.

Signed-off-by: Keegan Callin <kc@kcallin.net>
Bug: https://github.com/eclipse/mosquitto/issues/189

### DIFF
--- a/src/persist.c
+++ b/src/persist.c
@@ -350,9 +350,6 @@ int mqtt3_db_backup(struct mosquitto_db *db, bool shutdown)
 	char err[256];
 	char *outfile = NULL;
 	int len;
-#ifndef WIN32
-	int dir_fd;
-#endif
 
 	if(!db || !db->config || !db->config->persistence_filepath) return MOSQ_ERR_INVAL;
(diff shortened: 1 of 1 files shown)


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=189

kcallin:
Initial tests show that on Linux 4.1 persistence is working as well as it did before the patch.  I have two boards in reboot rigs to practically verify that the new (more atomic) write procedure doesn't break as far as we can test.  The practical testing will take another ~2 weeks to complete.

I think that the change should be merged if there's any risk of conflicts with other pending changes in the next two weeks.  If there's no risk then let's wait for the outcome of the practical tests.

kcallin:
Mechanical hardware testing is complete.  The test rig completed 15k power cycles without failure.  Formerly we would see failures in ~1-2k cycles.  Recommend merging.

ralight:
Thanks for this, I don't seem to have had a notification about your last message so only just saw it. I'll see about merging later today.

PR #210, opened 2016-07-12
Writing least significant byte of subscription QoS

tucic:
Signed-off-by: Milan Tucic <tucic.milan@gmail.com>


tucic:
I had a problem when restoring subscriptions for the persistence file on a MIPS platform. Stripped log file: [mosquitto-log.txt](https://github.com/eclipse/mosquitto/files/359044/mosquitto-log.txt)
Subscriptions are always written with qos=0, because int pointer is combined with sizeof(uint8_t). Issue related to this affects usage on big endians, when storing client's subscriptions (clean session=false).
I've checked the rest of persist.c and I think that there is no similar error present.

ralight:
Good catch, thanks very much.

Issue #107, opened 2016-03-15
Provide pkg-config library file


ralight:
migrated from Bugzilla [#433243](https://bugs.eclipse.org/bugs/show_bug.cgi?id=433243)
status UNCONFIRMED severity _enhancement_ in component _Mosquitto_ for _---_
Reported in version _unspecified_ on platform _All_
Assigned to: Roger Light

On 2014-04-22 15:02:31 -0400, Nicholas Humfrey wrote:

> Please can you provide and install libmosquitto.pc pkg-config file to make it easier to detect libmosquitto.

jensschroer:
@ralight I am happy to try to provide a PR for this but the debian branch seems to be broken and out of sync with debian upstream according https://qa.debian.org/cgi-bin/vcswatch?package=mosquitto 
Appreciate any suggestions on how I could help with this.

ralight:
This was completed for the Debian packaging a while back.

PR #211, opened 2016-07-12
Add pkg-config files to CMake build

simark:
This patch adds pkg-config files for libmosquitto and libmosquittopp,
and installs them through CMake.

Refs #107


giraldeau:
I tested this patch and confirm that it works as advertised.
```
$ find temp/ -type f | grep pkgconfig 
temp/lib/pkgconfig/libmosquittopp.pc
temp/lib/pkgconfig/libmosquitto.pc
```

simark:
Closing, because I had not read the contributing guidelines.  I'll re-open another pull request targeting the `develop` branch.

simark:
Please see #213 .

PR #213, opened 2016-07-12
Add pkg-config files to CMake build

simark:
This patch adds pkg-config files for libmosquitto and libmosquittopp,
and installs them through CMake.

Refs #107


ralight:
Thanks very much for this, sorry for the delay processing it.

PR #217, opened 2016-07-18
mosquitto_sub extra verbose option

nicopernas:
This very simple patch adds an extra-verbose option to mosquitto_sub that prints out the timestamp of when a message was received. 

I've dealt with this in the past by piping mosquitto_sub's output into a Perl script to append a timestamp to each line (message) and I thought it would be a good idea to have it in as part of the client.

Nico.


nicopernas:
Ok, I seem to have messed up my branches. I'll close this one and open a new one. Apologises for the spam

PR #218, opened 2016-07-18
Add extra-verbose option to mosquitto_sub

nicopernas:
This very simple patch adds an extra-verbose option to mosquitto_sub that prints out the timestamp of when a message was received.

I've dealt with this in the past by piping mosquitto_sub's output into a Perl script to append a timestamp to each line (message) and I thought it would be a good idea to have it in as part of the client.

Nico.

### DIFF
--- a/client/client_shared.c
+++ b/client/client_shared.c
@@ -749,6 +749,11 @@ int client_config_line_proc(struct mosq_config *cfg, int pub_or_sub, int argc, c
 				goto unknown_option;
 			}
 			cfg->verbose = 1;
+		}else if(!strcmp(argv[i], "-vv") || !strcmp(argv[i], "--extra-verbose")){
+			if(pub_or_sub == CLIENT_PUB){
+				goto unknown_option;
+			}
+			cfg->verbose = 2;
 		}else if(!strcmp(argv[i], "-x")){
 			if(pub_or_sub == CLIENT_PUB){
 				goto unknown_option;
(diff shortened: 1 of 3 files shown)


ralight:
Thanks for this, but I'm going to close it without merging it. The reason is that it has inspired me to a better solution which is a format option to allow any format a user desires.

I've implemented this as %t: topic, %l topic length, %q: qos, %m: mid, %p: payload as a string, %x: payload bytes in hex, %r: retain flag. Along side that, there is %TX, where the 'X' is passed to strftime as %X to allow any of the time formats allowed with strftime.

For example for an output format of "RFC3339 date : topic : payload in hex"

    mosquitto_sub -F '%TY-%Tm-%TdT%TH:%TM:%TS%Tz : %t : %x' -t '#'

ralight:
Would that suit you?

nicopernas:
Sorry for the delayed response. Yes that'd be perfect too! I thought about allowing a date format to be passed in so the user could select the way it wanted to see the date but I didn't really have time to implement it. But your solution would actually cover it! 

Thanks,
Nico.

ralight:
I'm closing this because the code i've mentioned has been committed.

SurenMJ:
Hi @ralight ,
On mosquitto_sub version 1.4.14 i'm getting "Error: Unknown option '-F'.". Could you please help on this?
Thanks,
Surendra

PR #220, opened 2016-07-19
About the codes style of mosquitto.

zhengshuxin:
I saw the codes of mosquitto, and feel these codes are so ugly. Can you rewrite them with a good style?

zsx


jpmens:
@zhengshuxin what exactly do you mean by "good style"? Can you show an example?

Also, you've created this as a huge pull-request: was that your intention?

karlp:
I don't see any commits from you @zhengshuxin in this series.  It appear you have the wrong branch selected in your PR perhaps? It's easier to review and provide comments if the PR is clean and easy to follow.

zhengshuxin:
@jpmens so many codes in this project,  such as mqtt3_db_message_insert and other functions in database.c. You compare all the source codes of this project with the Linux kernel, Postfix, Redis, Nginx. And you may know which style is a good code style.
zsx

ralight:
Thanks for your input, but I've got no intention of changing code style. I can be quite confident that any other project you approached would have the same response.

It is worth considering that beauty/ugliness are very much personal choice.

It is *also* worth noting that the coding style used in mosquitto is very similar to that of the Linux kernel, just with less white space in places.

zhengshuxin:
Even though each project has its owen code style, but they don't look so bad like mosquitto. The classic project's codes are orderly arranged which make the reader pleasing to appreciate the project.

ralight:
There have been a handful of people who have said they find the code clean and easy to read. There is one person who has said they don't like it, so I think I'm winning overall.

Even ignoring that, you don't make any suggestions. "Make it so I like it" isn't a useful comment. I could for example:

* Put all code in a single file to make it so you know which file to look in
* Put every function or type definition in its own separate source and header file so it is easy to find the file for a particular function / type
* Run all of the code through `indent -l10` so that it works on very narrow terminals

I'm sure that this isn't what you mean though.

FWIW, 'code style' to me means the kind of thing you could modify with the `indent` command.

PR #221, opened 2016-07-19
Add TLS server name indication (SNI) support

BrianEricsonExosite:
Introduce mosquitto_tls_server_name_set API to allow clients to specify
TLS server name indication (SNI) when connecting to brokers and provide
"--server-name" option in mosquitto_sub/mosquitto_pub to support usage.

Signed-off-by: Brian Ericson brianericson@exosite.com

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -830,6 +830,17 @@ int config__read_file_core(struct mosquitto__config *config, bool reload, const 
 #else
 					log__printf(NULL, MOSQ_LOG_WARNING, "Warning: Bridge and/or TLS support not available.");
 #endif
+				}else if(!strcmp(token, "bridge_server_name")){
+#if defined(WITH_BRIDGE) && defined(WITH_TLS)
+					if(reload) continue; // FIXME
+					if(!cur_bridge){
+						log__printf(NULL, MOSQ_LOG_ERR, "Error: Invalid bridge configuration.");
+						return MOSQ_ERR_INVAL;
+					}
+					if(conf__parse_string(&token, "bridge_server_name", &cur_bridge->tls_server_name, saveptr)) return MOSQ_ERR_INVAL;
+#else
+					log__printf(NULL, MOSQ_LOG_WARNING, "Warning: Bridge and/or TLS support not available.");
+#endif
 				}else if(!strcmp(token, "bridge_protocol_version")){
 #ifdef WITH_BRIDGE
 					if(reload) continue; // FIXME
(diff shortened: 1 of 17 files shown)


karlp:
I haven't looked at the code, but isn't SNI the sort of thing you want all the time, on by default?  Is the --server-name option intended for when people used only an IP address for the host?

src/conf.c
```diff
@@ -830,6 +830,17 @@ int config__read_file_core(struct mosquitto__config *config, bool reload, const
 #else
 					log__printf(NULL, MOSQ_LOG_WARNING, "Warning: Bridge and/or TLS support not available.");
 #endif
+				}else if(!strcmp(token, "bridge_server_name")){
+#if defined(WITH_BRIDGE) && defined(WITH_TLS)
+					if(reload) continue; // FIXME
```
karlp: You should probably mention FIXME's remaining in the code :)

BrianEricsonExosite: This came as copy-n-paste -- most of the "if(reload) continue;" lines in the file have a "FIXME" comment.  I retained it as part of following the pattern.

BrianEricsonExosite:
SNI is TLS's equivalent to HTTP 1.1's host header (both exist to disambiguate the intended server when multiple hostnames resolve to the same IP).  Whereas HTTP 1.1 made the header mandatory, it's optional in TLS.  It seems reasonable to always assign the server hostname to whatever is passed as the hostname, but this doesn't appear to be common practice.

karlp:
Just because it's not common doesn't mean it's ideal.  Wouldn't it make sense to always set SNI to the host given, and use the command line flag to set it to something else and/or disable it?  We want _more_ SNI, not less in the world.

anthonyrisinger:
How can we move this forward? Always sending SNI (matching `-h HOSTNAME`) with the option to override (`--server-name SERVERNAME`) is the preferred path. If the patch does not match this logic, we can update. Explicitly overriding SNI is crucial for testing alternative/local builds without resorting to `/etc/hosts` nonsense.

ralight:
Yes I agree. I've been testing the addition of SNI on the client and it behaves erratically. I've not figured out what the cause of the problem is yet, but it seems as though more work is required.

anthonyrisinger:
@ralight do you have any other details about the erratic behavior? I'm able to allocate resources towards SNI support.

sandro97git:
Is somebody still working on this issue?

willcharlton:
SNI is a requirement for my organization. Please add support for this as I would like to use mosquitto in applications.

ralight:
This has now been added via another PR, so I'm closing this one. Thank you for taking the time to contribute.

willcharlton:
Cool! Can you direct me to the PR? This is great news 🤘😁

PR #226, opened 2016-07-26
Adding Dockerfiles to build versioned Docker images

daudetCA:
Dockerfiles have been added to create Docker
images for Mosquitto v1.4.4 and v1.4.8 using Alpine
Linux as a base.

Different directories have been added for each version
to decouple the configuration and entrypoint scripts.
Currently they contain the same content, but this could
change in the future.

Docker Hub will also need to pull from specific directories
when official docker images will be built and hosted on
Docker Hub.

Signed-off-by: David Audet <david.audet@ca.com>


docker/1.4.4/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM alpine:3.3
+MAINTAINER David Audet <david.audet@ca.com>
+
+LABEL Description="Mosquitto MQTT Broker"
```
kartben: should be: "Eclipse Mosquitto MQTT Broker"

docker/1.4.4/README.md
```diff
@@ -0,0 +1,23 @@
+#Mosquitto v1.4.4 Docker Image
```
kartben: Eclipse Mosquitto

jpmens:
What @kartben says re the name "Mosquitto" applies further down in the commits for your 1.4.8 files as well, I assume.

I don't understand why you've duplicated things for Mosquitto, ahem, Eclipse Mosquitto 1.4.4 and 1.4.8; why not stick to the latest version?

daudetCA:
I was thinking it would be nice to have specific versions of Eclipse Mosquitto available via a tag in Docker Hub. For instance if someone wanted to use version 1.4.4 they could pull via eclipse-mosquitto:1.4.4. If another person wanted to use the latest, they could pull eclipse-mosquitto:latest from Docker Hub. In order to provide these different versioned images, we need specific Dockerfiles for each.

docker/1.4.4/README.md
daudetCA: I will update. Thanks!

docker/1.4.4/Dockerfile
daudetCA: I will update. Thanks!

karlp: Why? In the long form description, sure, but why should it be in the top level label?

docker/1.4.4/config/README.md
```diff
@@ -0,0 +1,4 @@
+# Configuration
+
+This directory is mounted as a volume for the docker container running Mosquitto.
+
```
karlp: Possibly here you could describe what you have changed in the configuration from the mosquitto defaults?

docker/1.4.4/config/mosquitto.conf.default
```diff
@@ -0,0 +1,819 @@
+# Config file for mosquitto
+#
```
karlp: This is a clone of the mosquitto.conf.example file right?  Why does it need to be a copy here?  It makes it far more work to update any config options if you have to keep this file up todate too.

docker/1.4.8/config/README.md
```diff
@@ -0,0 +1,4 @@
+# Configuration
+
+This directory is mounted as a volume for the docker container running Mosquitto.
+
```
karlp: So, this single pull adds _two_ docker images? for two different out of date versions? With all the files duplicated?

docker/1.4.4/Dockerfile
kartben: @karlp "Mosquitto" is a trademark of the Eclipse Foundation, and the trademark guidelines dictate that the first mention of the project must use its formal name prefixed with Eclipse.
You can read more at https://www.eclipse.org/projects/handbook/index.php#trademarks

kartben:
@jpmens We only really require that the first occurence of a project's name is properly attributed. We don't want nor need people to go crazy and clutter their websites, documentations, etc with things like "Eclipse® Mosquitto™" all over the place :)

docker/1.4.4/config/README.md
daudetCA: Sure, I can do that.

docker/1.4.4/config/mosquitto.conf.default
daudetCA: I guess it doesn't need to be here at all. My intent was to have an example config that could be changed if needed. I will remove these files.

docker/1.4.8/config/README.md
daudetCA: I have created different directories for each mosquitto version since the Dockerfile will be different for each version. This also gives flexibility in case other files need to be added for different versions of Mosquitto down the line. In order to have an official image in Docker Hub, you need to have a specific Dockerfile available that will be used to build that specific image (i.e. mosquitto-1.4.8). I am trying to follow the Docker Hub conventions that I am seeing with other images. We could just have 1 latest and greatest image, but I would prefer to allow older Mosquitto versions to be made available as well.

ralight:
Is there anything special about 1.4.4 and 1.4.8?

daudetCA:
There is nothing special about 1.4.4 and 1.4.8 other than those are the two versions that are in Alpine Linux's package repo for Alpine Linux version 3.3 and 3.4 which I am using as a base image.

PR #235, opened 2016-08-03
Fix missing context->listener for websocket client

PierreF:
The context associated with websocket client didn't had listener filled, which caused use_username_as_clientid to be ignored.

I've tested with the following config:
```
listener 1883
use_username_as_clientid true

listener 1880
protocol websockets
use_username_as_clientid true
```

When connecting to websocket as myuser:
```
#!/usr/bin/python
import paho.mqtt.client
c=paho.mqtt.client.Client(transport='websockets')
c.username_pw_set('myuser', 'secret')
c.connect('localhost', port=1880)
```

The clientid is NOT myuser:
```
1470235969: New client connected from 127.0.0.1 as paho/6CD3326A41CD4D51D9 (c1, k60, u'myuser').
```

### DIFF
--- a/src/mosquitto.c
+++ b/src/mosquitto.c
@@ -346,7 +346,7 @@ int main(int argc, char *argv[])
 			}
 		}else if(config.listeners[i].protocol == mp_websockets){
 #ifdef WITH_WEBSOCKETS
-			config.listeners[i].ws_context = mosq_websockets_init(&config.listeners[i], config.websockets_log_level);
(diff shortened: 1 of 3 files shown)


ralight:
Good catch.

Your patch looks good, but could you please base it off the fixes branch and you can use `_mosquitto_get_db()` to simplify the code quite a bit.

PierreF:
Created PR #239 which is based on fixes branch

PR #239, opened 2016-08-09
Fix missing context->listener for websocket client

PierreF:
The context associated with websocket client didn't had listener filled, which caused use_username_as_clientid to be ignored.

This PR replace PR #235


PierreF:
I've not used _mosquitto_get_db(), because the function already retrieved the db using &int_db.

ralight:
Thanks for your contribution.

PR #255, opened 2016-08-29
Publish topic with client connection state when client connects/disconnects.

jburhenn:
This change publishes a topic ($SYS/broker/clients/connection/_clientID_/state) with the client connection state whenever a client connects or disconnects. This is similar to the $SYS/broker/connection/_clientID_/state topic that is published for a bridge connection and allows for a simple way determine which clients are currently connected to the broker.

### DIFF
--- a/src/handle_connect.c
+++ b/src/handle_connect.c
@@ -569,6 +569,15 @@ int handle__connect(struct mosquitto_db *db, struct mosquitto *context)
 			}else{
 				log__printf(NULL, MOSQ_LOG_NOTICE, "New client connected from %s as %s (c%d, k%d).", context->address, client_id, clean_session, context->keepalive);
 			}
+			int notification_topic_len = strlen(client_id)+strlen("$SYS/broker/clients/connection//state");
+			char* notification_topic = mosquitto__malloc(sizeof(char)*(notification_topic_len+1));
+			if(notification_topic) {
+				log__printf(NULL, MOSQ_LOG_DEBUG, "Publish %s connection.", client_id);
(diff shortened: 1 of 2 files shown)


karlp:
Unlike bridges, which you can't manage yourself, this is something you can manage entirely in your application space.

jburhenn:
Do you mean by having clients publish to a topic on connect and use Last Will and Testament messages? Or is there another way? In our use case we can't be sure users connecting to our broker would send the correct connect & LWT messages so we were looking for a way to monitor client connections from within the broker itself.

karlp:
Yes, that's what I mean.  If they're not conforming clients, you block them from connecting to your service :)

jburhenn:
Okay, thanks for your feedback. We may look into doing it that way.

karlp:
Bear in mind, I'm just some dude on the internet, not a mosquitto developer or committer.  But the fact that this _can_ be implemented completely in client code is why it's not part of the standard, nor a standard feature.

jburhenn:
Ah, okay, I thought you were a mosquitto developer. Still, thanks for your input.

ralight:
Hi @jburhenn.

Thanks for your contribution. I'm inclined to agree with @karlp though, this is something that ought to be dealt with client side.

I'm also pretty nervous about adding things where the broker is automatically publishing potentially 10s of thousands of topics.

timoline:
+1 for client connection state by broker

jake2009:
Only the broker know clients state . if client offline of power off ,client will not send any message to broker.
so I need this topic .

ralight:
@jake2009 The code is there for you to use if you want, it's not going into the broker as it stands. It's possible a future plugin interface might let you do this, but please don't wait for that to happen.

Issue #252, opened 2016-08-28
NetBSD support


amoibos:
Hello,
it would be nice if could integrate some patches to recognize NetBSD in your next release. 

--- lib/CMakeLists.txt.orig 2016-06-02 21:20:22.000000000 +0000
+++ lib/CMakeLists.txt  2016-08-28 14:00:33.700421980 +0000
@@ -43,9 +43,9 @@

 set (LIBRARIES ${OPENSSL_LIBRARIES} ${PTHREAD_LIBRARIES})

-if (UNIX AND NOT APPLE)
+if (UNIX AND NOT APPLE AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "NetBSD")
    set (LIBRARIES ${LIBRARIES} rt)
-endif (UNIX AND NOT APPLE)
+endif (UNIX AND NOT APPLE AND NOT ${CMAKE_SYSTEM_NAME} MATCHES "NetBSD")

 if (WIN32)
    set (LIBRARIES ${LIBRARIES} ws2_32)

--- lib/memory_mosq.h.orig  2016-06-02 21:20:22.000000000 +0000
+++ lib/memory_mosq.h   2016-08-28 14:30:34.000000000 +0000
@@ -20,7 +20,7 @@
 #include <stdio.h>
 #include <sys/types.h>

-#if defined(WITH_MEMORY_TRACKING) && defined(WITH_BROKER) && !defined(WIN32) && !defined(**SYMBIAN32**) && !defined(**ANDROID**) && !defined(**UCLIBC**) && !defined(**OpenBSD**)
+#if defined(WITH_MEMORY_TRACKING) && defined(WITH_BROKER) && !defined(WIN32) && !defined(**SYMBIAN32**) && !defined(**ANDROID**) && !defined(**UCLIBC**) && !defined(**OpenBSD**) && !defined(**NetBSD**)
 #define REAL_WITH_MEMORY_TRACKING
 #endif

--- src/CMakeLists.txt.orig 2016-06-02 21:20:22.000000000 +0000
+++ src/CMakeLists.txt  2016-08-28 14:13:03.999103447 +0000
@@ -85,6 +85,8 @@
 if (UNIX)
    if (APPLE)
        set (MOSQ_LIBS ${MOSQ_LIBS} dl m)
-   elseif (${CMAKE_SYSTEM_NAME} MATCHES "NetBSD")
-       set (MOSQ_LIBS ${MOSQ_LIBS} m)

[...]

PR #258, opened 2016-09-02
Add NetBSD Support

amoibos:
Signed-Off-By: Daniel Ölschlegel amoibos@gmail.com

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -115,7 +115,7 @@ LIB_LDFLAGS:=${LDFLAGS}
 BROKER_CFLAGS:=${LIB_CFLAGS} ${CPPFLAGS} -DVERSION="\"${VERSION}\"" -DTIMESTAMP="\"${TIMESTAMP}\"" -DWITH_BROKER
 CLIENT_CFLAGS:=${CFLAGS} ${CPPFLAGS} -I../lib -DVERSION="\"${VERSION}\""
 
-ifneq ($(or $(findstring $(UNAME),FreeBSD), $(findstring $(UNAME),OpenBSD)),)
+ifneq ($(or $(findstring $(UNAME),FreeBSD), $(findstring $(UNAME),OpenBSD), $(findstring $(UNAME),NetBSD)),)
 	BROKER_LIBS:=-lm
 else
 	BROKER_LIBS:=-ldl -lm
--- a/lib/memory_mosq.h
+++ b/lib/memory_mosq.h
@@ -20,7 +20,7 @@ Contributors:
 #include <stdio.h>
 #include <sys/types.h>
 
-#if defined(WITH_MEMORY_TRACKING) && defined(WITH_BROKER) && !defined(WIN32) && !defined(__SYMBIAN32__) && !defined(__ANDROID__) && !defined(__UCLIBC__) && !defined(__OpenBSD__)
+#if defined(WITH_MEMORY_TRACKING) && defined(WITH_BROKER) && !defined(WIN32) && !defined(__SYMBIAN32__) && !defined(__ANDROID__) && !defined(__UCLIBC__) && !defined(__OpenBSD__) && !defined(__NetBSD__)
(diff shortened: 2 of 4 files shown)


karlp:
in general, (I'm not a developer or committer here) your commit messages should have somethign useful in them, like, "Support building on NetBSD" rather than the incredibly vague "update file X"  "Update file x" is immediately obvious from the diffstat, what's more interesting in the commit comment is the why and how the changes made were done.

config.mk
```diff
@@ -115,7 +115,7 @@ LIB_LDFLAGS:=${LDFLAGS}
 BROKER_CFLAGS:=${LIB_CFLAGS} ${CPPFLAGS} -DVERSION="\"${VERSION}\"" -DTIMESTAMP="\"${TIMESTAMP}\"" -DWITH_BROKER
 CLIENT_CFLAGS:=${CFLAGS} ${CPPFLAGS} -I../lib -DVERSION="\"${VERSION}\""
 
-ifneq ($(or $(findstring $(UNAME),FreeBSD), $(findstring $(UNAME),OpenBSD)),)
+ifneq ($(or $(findstring $(UNAME),FreeBSD), $(findstring $(UNAME),OpenBSD), $(findstring $(UNAME),NetBSD)),)
```
karlp: Perhaps it's time to just make this look for "BSD" ?  dragonfly, miros, trusted, pc, et al?

lib/memory_mosq.h
```diff
@@ -20,7 +20,7 @@ and the Eclipse Distribution License is available at
 #include <stdio.h>
 #include <sys/types.h>
 
-#if defined(WITH_MEMORY_TRACKING) && defined(WITH_BROKER) && !defined(WIN32) && !defined(__SYMBIAN32__) && !defined(__ANDROID__) && !defined(__UCLIBC__) && !defined(__OpenBSD__)
+#if defined(WITH_MEMORY_TRACKING) && defined(WITH_BROKER) && !defined(WIN32) && !defined(__SYMBIAN32__) && !defined(__ANDROID__) && !defined(__UCLIBC__) && !defined(__OpenBSD__) && !defined(__NetBSD__)
```
karlp: so freebsd supports memory tracking, but not openbsd or netbsd.  Perhaps this is really a check on libc versions?  (much like the existing uclibc check?)

lib/net_mosq.c
```diff
@@ -34,7 +34,7 @@ and the Eclipse Distribution License is available at
 #include <sys/endian.h>
 #endif
 
-#ifdef __FreeBSD__
+#if defined(__FreeBSD__) || defined(__NetBSD__)
```
karlp: Possibly, these checks should go to the config.mk, and set something like, "HAS_NETINET_IN_H" instead, rather than modifying all the files individually?

config.mk
amoibos: My intention was to add the support for my OS and reduce the size of the patchset on Pkgsrc,, not to refactoring foreign code. For Dragonfly, NetBSD, OpenBSD, FreeBSD and many derivates this could work. But not for Bitrig.

lib/net_mosq.c
amoibos: Maybe but when i look on the patches in Ports and Pkgsrc, is this maybe a smaller problem. For example NetBSD doesn't support long options for install, this should also the case for all BSDs.

ralight:
Thanks for these fixes. I've just pushed a slightly different version of the changes based on some of Karl's suggestions, so I'm going to close this.

You can avoid the install long options using `make WITH_STRIP=yes STRIP_OPTS=-s`. The default now is to not strip anyway.

gdt:
Thanks for integrating this.  I just updated pkgsrc-wip/mosquitto to 1.5.1, and was able to drop a lot of patches.   Also you can close #252 as a dup of this.

PR #260, opened 2016-09-05
websockets: fix compatibility with older lws versions

karlp:
patch is against the fixes branch, this was a regression from 1.4.9 to 1.4.10.
See also https://github.com/openwrt/packages/issues/3127

----
In 1.3, 1.4 and 1.5, the function was "libwebsockets_get_protocol" not
"libwebsocket_get_protocol"  While the #define name doesn't matter on
newer libwebsockets, where it redirects to lws_get_protocol, the naming
is critical for older versions.

Fixes: 477cd3e39911 (Fix missing context->listener for websocket client)

Signed-off-by: Karl Palsson <karlp@etactica.com>


ralight:
Thanks.

As a heads up, I'm thinking about removing support for lws <2.x in mosquitto 1.5. Is that going to be a problem?

karlp:
Shouldn't be for me, this was just problematic that a point release (.9 to .10) broke on an prior working LWS version.

trunk openwrt/lede has moved on to LWS 2 already now, this only mattered on the old stable release branch

Issue #270, opened 2016-09-16
Maximum connections for websockets listener


tucic:
This is not checked before accepting new clients for websockets listeners. [websockets.c:182](https://github.com/eclipse/mosquitto/blob/master/src/websockets.c#L182)

PierreF:
Fixed by PR #271

PR #271, opened 2016-09-16
Maximum connections for websockets listener

tucic:
Check current number of connections before accepting new websockets clients.

Signed-off-by: tucic <tucic.milan@gmail.com>


tucic:
Bug: #270

ralight:
Thanks very much, this looks fine.

Issue #311, opened 2016-11-09
omit update symbol in websockets.c


oxbambooxo:
build from source, set WITH_LIBWEBSOCKETS=yes, get the error 

    undefined reference to `libwebsocket_get_protocol'

in [websockets.c#L185](https://github.com/eclipse/mosquitto/blob/v1.4.10/src/websockets.c#L185)

using old define 'libwebsocket_get_protocol', should be 'libwebsockets_get_protocol'

karlp:
Already fixed in https://github.com/eclipse/mosquitto/commit/03d4b8c27029da25dd350df3ca86dd7a6f1393fe

oxbambooxo:
oh, sorry, i'm download the master branch, it's ahead of master branch.

PR #314, opened 2016-11-20
	modified:   mosquitto_broker.h

piebat:
modified:   websockets.c and mosquitto_broker.h
Fix a typo error:  libwebsockets_get_protocol() and Not  libwebsocket_get_protocol
Now compiling with WITH_WEBSOCKETS:=yes, doesn't give error.

### DIFF
--- a/src/mosquitto_broker.h
+++ b/src/mosquitto_broker.h
@@ -31,7 +31,7 @@ Contributors:
 #    define libwebsocket_write(A, B, C, D) lws_write((A), (B), (C), (D))
 #    define libwebsocket_get_socket_fd(A) lws_get_socket_fd((A))
 #    define libwebsockets_return_http_status(A, B, C, D) lws_return_http_status((B), (C), (D))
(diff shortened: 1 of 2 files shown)


karlp:
Already fixed, see https://github.com/eclipse/mosquitto/pull/260 and https://github.com/eclipse/mosquitto/issues/311

ralight:
This is already fixed as @karlp says. Thanks for your contribution anyway!

FernandoGarcia:
Hi!

@ralight @karlp Why you are talking it's fixed if it is not available in master branch?

https://github.com/eclipse/mosquitto/blob/master/src/mosquitto_broker.h#L37

Best regards.

karlp:
@FernandoGarcia mosquitto uses the "fixes" and "develop" branch, it doesn't use the "master" branch the same way as you might expect.  (for better or worse)  fixes is stable, develop is what you might consider comparable to "master" in some other projects.

FernandoGarcia:
@karlp Thanks!

PR #320, opened 2016-11-23
Add support for local only bridge notifications

### LINKED ISSUE #233: Enchancement: Bridge notifications only locally
Would it be possible to introduce a configuration that would publish bridge state only locally (0 | 1). A remote broker does not allow retained messages to be published. However the notifications can be used on a device. E.g. I was thinking to make Raspberry Pi senseHAT App that would monitor the notification and then make it into LED light ( green | red ) based on state.

hardillb:
This update adds an option to only publishes bridge
notification messages to the local side of the bridge.

It adds a config file option called notifications_local_only
that accepts a boolean value, defaults to false to be
consistent with existing behaviour.

Fixes #233

Signed-off-by: Ben Hardill <hardillb@uk.ibm.com>


hardillb:
Hi @ralight any chance we can get this considered for the feature list of the next release? Thanks.

ralight:
This looks good to me, but could you please rebase against the develop branch?

hardillb:
Sure, I've made the changes tonight, but I need to give them a proper test before I update the pull request. Should have it done at the latest on Monday

PR #338, opened 2016-12-21
Update Dockerfiles to not save apk cache

daudetCA:
When installing the Mosquitto package from Alpine Linux, we don't need
to save the latest package index to the cache. We are specifically requesting
the latest package index each time we install Mosquitto and not using the local
cache at all (--no-cache).


ralight:
Thanks very much.

PR #340, opened 2016-12-26
Add 'mosquitto_log_vprintf' function for plugins

sandynomad:
The existing logging interface doesn't support non variable (ie: va_list) arguments.


sandynomad:
Please ignore.

PR #344, opened 2016-12-29
Added support for asynchronous DNS-query.

### LINKED ISSUE #41: When internet is down, bridge connection DNS request blocks mosquitto main loop.
migrated from Bugzilla [#472600](https://bugs.eclipse.org/bugs/show_bug.cgi?id=472600)
status UNCONFIRMED severity _normal_ in component _Mosquitto_ for _1.4_
Reported in version _unspecified_ on platform _PC_
Assigned to: Roger Light

On 2015-07-14 05:34:52 -0400, jsaak jsaak wrote:

> An async dns request should be used.
> As a workaround i put ip address instead of domain name.

kariok:
This update adds an option to use asynchronous DNS-query on Linux-environment.
It runs async-query on background and gives time for mosquitto to serve local clients.

Fixes #41

Signed-off-by: Kari Keinanen <kari.keinanen@kone.com>

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -80,6 +80,9 @@ WITH_DOCS:=yes
 # Build with client support for SOCK5 proxy.
 WITH_SOCKS:=yes
 
+# Build with asynchronous DNS queries (use getaddrinfo_a() instead of getaddrinfo()).
+WITH_ASYNC_DNS:=no
+
 # Strip executables and shared libraries on install.
 WITH_STRIP:=no
 
@@ -194,6 +197,13 @@ endif
 ifeq ($(WITH_SOCKS),yes)
 	LIB_CFLAGS:=$(LIB_CFLAGS) -DWITH_SOCKS
 	CLIENT_CFLAGS:=$(CLIENT_CFLAGS) -DWITH_SOCKS
+endif
+
+ifeq ($(WITH_ASYNC_DNS),yes)
+	ifeq ($(UNAME),Linux)
+		BROKER_CFLAGS:=$(BROKER_CFLAGS) -DWITH_ASYNC_DNS -D_GNU_SOURCE
+		BROKER_LIBS:=$(BROKER_LIBS) -lanl
+	endif
 endif
 
 ifeq ($(WITH_UUID),yes)
--- a/lib/net_mosq.c
+++ b/lib/net_mosq.c
@@ -19,6 +19,10 @@ Contributors:
 #include <fcntl.h>
 #include <stdio.h>
 #include <string.h>
+#ifdef WITH_ASYNC_DNS
+#include <signal.h>
+#include <stdlib.h>
+#endif
 #ifndef WIN32
 #include <netdb.h>
 #include <sys/socket.h>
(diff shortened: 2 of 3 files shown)


hardillb:
While I like this as a much more simple solution the one I've been looking at it does have a number of problems:

1. I don't thing this will build on Windows as there doesn't look to be a `getaddrinfo_a` function in the standard libraries.
2. It does mean that it will always wait for one bridge timeout period before connecting as the first time round the loop it "fails" to do the look up then doesn't check for a result until the next time it tries to bring the bridge up.

The second problem is possibly acceptable, but I fear the first issue will be more of a problem as we need to make this work on all platforms.

kariok:
This is kind of important for us (we need local MQTT-communication while mosquitto is resolving bridge connection).

1. I think there is DnsQueryEx function in Windows-DNS-API

2. I couldn't find easy solution for triggering another try when query completes (without changing the logic too much)

Another possible solution could be to use separate thread for background dns-query.

hardillb:
1. The c-ares library offers cross platform support for async DNS look ups, it is already used in the mosquitto clients and is what I've been looking at using. I think this direction is best as we don't want to have to maintain 2 separate versions of the code for Linux/Windows.

 2. Unfortunately I think that a major restructure of bridge connection code is the only way to do this.

ralight:
Thanks for looking into this, it's definitely a bugbear.

Am I right in saying that with this implementation it's only possible to look up one bridge address at once? So @hardillb's point about the delay would get compounded if there were multiple bridges?

As far as this goes, ultimately it needs to get fixed properly - which is most likely as part of a move to e.g. libuv for the network loop which also provides async lookups. Until that point a temporary solution wouldn't be so bad.

@hardillb : as things stand we don't distribute c-ares on Windows anyway, so that isn't a magic bullet. To be honest, c-ares seems dead so I'd like to drop it at some point. It won't be used by default in 1.5.

hardillb:
c-ares Dev seams to have moved to github and there have been updates in the last month https://github.com/c-ares/c-ares

But a move to libuv would be a interesting move (if a large change)

kariok:
"only possible to look up one bridge address at once?"
Yes. If wanted to resolve multiple bridges at once, resolving should be done somewhere else (current network handling code only has one address).

"c-ares (or similar)"
This might be the long term solution but maybe needs more refactoring.

But in the mean while, can't we have some solution preventing mosquitto stuck for ~20seconds.
In that time, which is the current situation, local MQTT client communication is not possible at all.

hardillb:
If you remove dnsmasq the 20seconds drops to 5 seconds. This should have been passed to you via the IBM account team before Christmas.

I'd missed that it only brings up one bridge at a time, to me that is another vote for having to do the full refactor.

kariok:
"20seconds drops to 5 seconds".
Mosquitto retries to resolve connection every 30seconds, so when bridge connection is lost, mosquitto gets stuck for 5seconds in 30seconds periods.
Sorry, but this does not sound good.

hardillb:
I was just pointing out that the 20 seconds can be substantially reduced, 5 seconds every 30 is a lot better than 20 seconds every 30. It's still not perfect, but in testing it did allow clients to remain connected.

kariok:
Any plans/time estimation for doing refactoring/another solution ?

ralight:
I'm working on it with a plan for Linux support only for 1.4.11.

kariok:
So, 1.4.11 (Linux version) will contain refactored solution and no DNS-stuck at all, right ?

ralight:
Yes, that's the plan.

kariok:
Great, looking forward to get it in use.

ralight:
I've pushed the in-progress work to the fixes branch if you would like to give it a try. I'm not in a position to test it with bad DNS right now, but it works with working DNS.

kariok:
Now the local communication does not stuck when mosquitto is resolving DNS.
Only run some simple tests, but looks good.

ralight:
I've found bugs in testing, so it's not quite ready yet unfortunately.

hardillb:
I'm away this week but I can share (and test in) my emulated test environment when I get back at the weekend

ralight:
That'd be good, thanks. It works for my tests now, but I've still not done anything with broken dns :)

It also needs to check whether getaddrinfo_a is actually available.

ralight:
This now needs `make WITH_ADNS=yes` to compile support in if you've using makefiles rather than cmake. Seems the best option for now.

hardillb:
which branch are these updates on?

ralight:
I think this is now sorted for Linux glibc devices, so I'm going to close this PR.

PR #345, opened 2017-01-02
Fixes -Wreserved-id-macro in mosquitto.h and mosquittopp.h.

jbreitbart:
In C and C++ identifiers starting with an underscore are reserved. This
commit fixes the warning for the files included by other projects.

Signed-off-by: Jens Breitbart <jbreitbart@gmail.com>

### DIFF
--- a/lib/cpp/mosquittopp.h
+++ b/lib/cpp/mosquittopp.h
@@ -14,8 +14,8 @@ Contributors:
    Roger Light - initial implementation and documentation.
 */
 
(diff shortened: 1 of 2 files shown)


ralight:
This was already fixed every so slightly differently in the develop branch. I've made the same change in the fixes branch rather than taking this PR so I don't have to resolve conflicts later on.

I'm going to close this, thanks for your contribution anyway!

PR #360, opened 2017-01-24
Check for rt and pthread libraries before linking

fornwall:
This fixes building on Android which does not have separate librt or libpthread libraries.


ralight:
Thanks very much, I've merged this on the fixes branch instead:

ralight:
https://github.com/eclipse/mosquitto/commit/532273250c98568ad086807880a3f4a7e204a7d7

PR #361, opened 2017-01-24
Include <syslog.h> instead of <sys/syslog.h>

fornwall:
Using the standard <syslog.h> header instead of <sys/syslog.h> makes it consistent with other source files and fixes compilation on Android.

Signed-off-by: Fredrik Fornwall <fredrik@fornwall.net>

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -36,7 +36,7 @@ Contributors:
 #endif
 
 #if !defined(WIN32) && !defined(__CYGWIN__)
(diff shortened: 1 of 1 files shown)


ralight:
Thanks, not sure why this one was sys/syslog.h. I've merged it as

https://github.com/eclipse/mosquitto/commit/23b0891048ea017aded192da74df624b785392ce

PR #368, opened 2017-02-01
Log more TLS error details

pinkavaj:
Signed-off-by: Jiří Pinkava <j-pi@seznam.cz>

This adds more verbose logging for case when TLS error occures during connection. This is especially aimed in case where CN in server certificate does not match server hostname and 'insecure_tls' option is disabled (default).

Without this error logging messages client (eg. mosquitto_pub) says nothing useful and server only says "Internal SSL error".

This happens when someone creates certificates as described in mosquitto-tls and uses mosquitto server with default example configuration. I suppose this might happens quite often. Debuging this is pretty hard and requires to dwelling deep in source code.

### DIFF
--- a/lib/net_mosq.c
+++ b/lib/net_mosq.c
@@ -321,6 +321,9 @@ void net__print_ssl_error(struct mosquitto *mosq)
 int net__socket_connect_tls(struct mosquitto *mosq)
 {
 	int ret, err;
+	unsigned long e;
+	char ebuf[256];
+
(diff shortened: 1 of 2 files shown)


pinkavaj:
Waits for #386 to merge, then i will push -f branch with fixed build issue.

ralight:
I'm fully in support of better TLS logging, thanks very much for this. I've fixed the build error myself as well.

PR #378, opened 2017-02-10
Minor typo

stefannilsson:


ralight:
Fixed with a separate commit, thanks for spotting it.

PR #273, opened 2016-09-17
Update readme.md

YeGrantyuan:
fix Build Dependencies bug

PR #400, opened 2017-03-08
Fixed the incorrect define name

nickivanov:
for c-ares dependency


karlp:
And this itself is a duplicate of #273

ralight:
Thanks for reporting this, I've fixed it with a separate commit because it's a trivial patch.

PR #402, opened 2017-03-14
Fix compilation against musl libc.

stintel:
Commit e13af18ed9c527b8152697c9ede62f0e4e81bdd6 breaks mosquitto on
systems with musl libc, due to unconditional inclusion of -lanl on
Linux.

Commit 433ee5c4d6852b507b43eae9c9a2c9028c6b48e5 already added
conditional inclusion of -lanl, so just remove the unconditional
inclusion completely.

Signed-off-by: Stijn Tintel <stijn@linux-ipv6.be>

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -159,10 +159,6 @@ ifeq ($(UNAME),QNX)
 	LIB_LIBS:=$(LIB_LIBS) -lsocket
 endif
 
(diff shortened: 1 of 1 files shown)


ralight:
Thanks for this, I've fixed it with another PR and will be making a new release soon.

PR #414, opened 2017-04-01
send CONNACK to websocket client authorized failed

priest334:
Signed-off-by: priest334 <priest334@126.com>


priest334:
The modification runs ok with my test-case

priest334:
ECA passed

PR #415, opened 2017-04-03
Remove -lanl when WITH_ADNS is unset

ffontaine:
Do not add -lanl to BROKER_LIBS for all Linux builds.
Indeed, -lanl is only needed for getaddrinfo_a which is only used in
_mosquitto_try_connect_step1 when WITH_ADNS is set

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -159,10 +159,6 @@ ifeq ($(UNAME),QNX)
 	LIB_LIBS:=$(LIB_LIBS) -lsocket
 endif
 
-ifeq ($(UNAME),Linux)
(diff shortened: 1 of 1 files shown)


ralight:
Thank you for spotting this, and sorry for the delay in responding. I've pushed the commit to the fixes branch so I'm closing this PR.

https://github.com/eclipse/mosquitto/commit/439f39b5daba0de131cc0c28feacea851cf2c08e

PR #418, opened 2017-04-06
Add new feature: Apache Kafka bridge

mklostermann:
Mosquitto can be configured to forward messages to an Apache Kafka broker.
The configuration is desribed in mosquitto.conf.
The compiler flag WITH_KAFKA_BRIDGE needs to be set to use the new
feature. Currently, the Kafka bridge only ensures QoS 0.

Dependencies: librdkafka
Test dependencies: python-kafka

Signed-off-by: Markus Klostermann <mklostermann@student.tugraz.at>

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -35,7 +35,11 @@ WITH_THREADING:=yes
 # to connect to other brokers and subscribe/publish to topics. You probably
 # want to leave this included unless you want to save a very small amount of
 # memory size and CPU time.
-WITH_BRIDGE:=yes
+#WITH_BRIDGE:=yes
+
+# Comment out to remove Kafa support from the broker. This allows the broker
+# to connect to a Kafka cluster and publish messages to it.
+WITH_KAFKA_BRIDGE:=yes
 
 # Comment out to remove persistent database support from the broker. This
 # allows the broker to store retained messages and durable subscriptions to a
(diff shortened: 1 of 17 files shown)


mklostermann:
The build fails because of a missing dependency, librdkafka. I think a cannot do anything about it, someone else maybe?

config.mk
```diff
@@ -35,7 +35,11 @@ WITH_THREADING:=yes
 # to connect to other brokers and subscribe/publish to topics. You probably
 # want to leave this included unless you want to save a very small amount of
 # memory size and CPU time.
-WITH_BRIDGE:=yes
+#WITH_BRIDGE:=yes
```
karlp: Don't think you want to turn off normal bridging here

karlp:
Can you give some examples of why this is more preferable than kafka learning mqtt?  Or just a process that subcribes and sends to kafka itself?  Why would a clean contained mqtt broker want _one_ particular extra destination style?  This feels very out of scope to me, and virtually impossible to maintain.  Presumably the motivation is to do the forwarding "in process" rather than the perceived overhead of a second process?

config.mk
mklostermann: Both should work.

mklostermann:
The idea for the direct integration came up while I was working on my master's thesis. You are right, I thought it would be better than an additional process. This way I can get rid of additional processes and network connections. I think the advantages are similar to that of using Mosquitto's MQTT bridge feature.

To be honest, I did not understand the quality of service levels when this idea came to my mind. Today I would choose a different solution. Nevertheless, I still think this would be a valuable feature and wanted to share the code.
But I would also understand if you reject the pull request because of the reasons you mentioned in your comment.

ralight:
It's good to see people doing different things with mosquitto. I'm afraid I don't think that this is a good fit for the project, it's a bit too specific. I don't want to open up a can of worms by adding this.

PR #421, opened 2017-04-07
Maximize out buffer size

hollosigergely:
We stuck in a problem when using MQTT QoS 1 or 2 that if a connection drops the out queue is growing without a limit (consuming all the memory) and - it seems, I may be wrong - there is no way of noticing that the connection has dropped (in case of *mosquitto_loop_start*). We made a little change to specify the out queue size and passing an error in *mosquitto_publish* when the queue is full. If you consider this a good feature I can fit the interface to mosquitto guidelines (if you provide one :)).

### DIFF
--- a/lib/messages_mosq.c
+++ b/lib/messages_mosq.c
@@ -113,6 +113,7 @@ void mosquitto_message_free(struct mosquitto_message **message)
  * Returns:
  *	0 - to indicate an outgoing message can be started
  *	1 - to indicate that the outgoing message queue is full (inflight limit has been reached)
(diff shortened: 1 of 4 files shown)


ralight:
Sorry for the delay replying to you.  I like this idea, I admit it's a problem at the moment. From the quick look I've just had it looks broadly ok, but I have two comments for now:

Please rebase on the develop branch - that's the place for new features.
You'll need to add a new entry to lib/linker.version for the new function - it'll need a new section at the bottom MOSQ_1.5. If you're not sure, I can do it.

PR #428, opened 2017-04-21
Fix bridge->restart_t won't be reset

imZack:
If bridge use a non-existed PSK Identity or connection failed it will retry to connect continuously without any timeouts.

This issue is introduced by [Start of fix for [344]](https://github.com/eclipse/mosquitto/commit/e13af18ed9c527b8152697c9ede62f0e4e81bdd6#diff-3ca823489e2803bdd90f0774bbdd0c09L266).

Affected version: v1.4.11


ralight:
Thanks very much for this, I've added the patch to the fixes branch instead, so I'm closing the PR.

https://github.com/eclipse/mosquitto/commit/97572610c07955ffa8496cf1527023ad3f9e80ad

PR #436, opened 2017-05-03
Fix documentation concerning WITH_SRV flag

stefan-hudelmaier:
The flag `WITH_DNS_SRV` in readme.md had no effect. The actually used flag seems to be `WITH_SRV`.


karlp:
Duplicate of #400

ralight:
Thanks for reporting this, I've fixed it with a separate commit because it's a trivial patch.

PR #441, opened 2017-05-07
Out queue maximum size

hollosigergely:
I rebased my work on limiting the outgoing queue size onto the develop branch. I edited the linker.version file as well. Apart from these I introduced a new error constant because no other constant explains the situation exactly (NOMEM is used for memory allocation errors).


mgsteinkamp:
With the current setting, what happens if input > output? Does mosquitto start to toss old messages? Ignore new messages? Pop?

Issue #446, opened 2017-05-12
Prevent no unique connection bridge name


Tifaifai:
Hello,

In case of no unique connection bridge name in mosquitto.conf, mosquitto crash with a segmentation fault.

example:
```
connection test
address localhost:1884
topic # both testbrod1/ testbrod1/

connection test
address localhost:1884
topic # both testbrod2/ testbrod2/
```
We know that it's not a good practice (same name for connection) but the reaction of mosquitto can be better.
A first bad solution is prevent this bug with rename the bridge by default
in ```conf.c```, line 928 
```
cur_bridge = &(config->bridges[config->bridge_count-1]);
memset(cur_bridge, 0, sizeof(struct _mqtt3_bridge));
cur_bridge->name = _mosquitto_strdup(token);
if(!cur_bridge->name){
	_mosquitto_log_printf(NULL, MOSQ_LOG_ERR, "Error: Out of memory.");
	return MOSQ_ERR_NOMEM;
}
```
replace by:
```
cur_bridge = &(config->bridges[config->bridge_count-1]);
memset(cur_bridge, 0, sizeof(struct _mqtt3_bridge));

sprintf(token, "%s%d", token, config->bridge_count);

cur_bridge->name = _mosquitto_strdup(token);
if(!cur_bridge->name){
	_mosquitto_log_printf(NULL, MOSQ_LOG_ERR, "Error: Out of memory.");
	return MOSQ_ERR_NOMEM;
}
```

Have you a better solution for prevent segmentation fault and crash ? because we don't understand this problem (no use ```bridges[i]->name``` after).

THX

Tifaifai:
Hello,
We can propose you a new better solution :)
In ```src/bridge.c```in function ```int mqtt3_bridge_new```
Replace:
```
if(!bridge->remote_clientid){
	if(!gethostname(hostname, 256)){
		len = strlen(hostname) + strlen(bridge->name) + 2;
		id = _mosquitto_malloc(len);
		if(!id){
			return MOSQ_ERR_NOMEM;
		}
		snprintf(id, len, "%s.%s", hostname, bridge->name);
	}else{
		return 1;
	}
bridge->remote_clientid = id;
```
by:
```
if(!bridge->remote_clientid){
	if(!gethostname(hostname, 256)){
		len = strlen(hostname) + strlen(bridge->name) + 2 + 1 + snprintf(NULL, 0, "%d", db->bridge_count+1);
		id = _mosquitto_malloc(len);
[...]

PR #447, opened 2017-05-12
Prevent no unique bridge name (connection option)

Tifaifai:
After this discussion https://github.com/eclipse/mosquitto/issues/446, we propose you a little correction.

THX


ralight:
Thanks for this contribution, but I've made the fix in a different way so I'm closing this pull request.

PR #456, opened 2017-06-08
listener_allow_anonymous

kurt-vd:
This commit introduces a (per-listener) listener_allow_anonymous
option that controls what to do with anonymous connections.
For backward compatibility, this option is prefixed with 'listener_'
and the global 'allow_anonymous' is still in use, i.e. anonymous
connections are allowed if any of allow_anonymous and
listener_allow_anonymous are true.

Signed-off-by: Kurt Van Dijck <dev.kurt@vandijck-laurijssen.be>


kurt-vd:
I'll prepare a new pull request that is based on a working develop branch ...

PR #467, opened 2017-06-20
Add unix domain socket listener.

toast-uz:
I added unix domain socket listener.
Because I use mosquitto with Nginx stream proxy on the same machine frequently, it will be better the communication between mosquitto and Nginx can support unix domain socket in order to reduce internal TCP port resources and optimize performance. Plz evaluate it. Thanks.
#USAGE: listener unix:<path_name>

Signed-off-by: Tatsuzo Osawa  (toast-uz) #

### DIFF
--- a/man/mosquitto.conf.5.xml
+++ b/man/mosquitto.conf.5.xml
@@ -700,6 +700,17 @@
 					</listitem>
 				</varlistentry>
 				<varlistentry>
+					<term><option>listener</option> <replaceable>unix:unix-path</replaceable></term>
+					<listitem>
+						<para>Listen for incoming internal connection on the
+							unix domain socket represented by unix-path.
+						</para>
+						<para>This can be applied both mqtt and websockets. To use websockets, this need libwebsockets 2.0 or later compiled with unix domain socket support option. Mosquitto must have proper privileges to handle the path.</para>
+						<para>This option may be specified multiple times. Mosquitto can create simultaniously both TCP and unix listener.</para>
+						<para>Not reloaded on reload signal.</para>
+					</listitem>
+				</varlistentry>
+				<varlistentry>
 					<term><option>max_connections</option> <replaceable>count</replaceable></term>
 					<listitem>
 						<para>Limit the total number of clients connected for
@@ -1522,3 +1533,4 @@ topic clients/total in 0 test/mosquitto/org $SYS/broker/
 		<para>Roger Light <email>roger@atchoo.org</email></para>
 	</refsect1>
 </refentry>
+
--- a/mosquitto.conf
+++ b/mosquitto.conf
@@ -291,6 +291,15 @@
 # Note that for a websockets listener it is not possible to bind to a host
 # name.
 # listener port-number [ip address/host name]
+#
+# If unix: directive and unix path is given, mosquitto will attempt
+# to bind the listener to an unix domain socket represented by the path.
+# This can be applied both mqtt and websockets. To use websockets,
+# this need libwebsockets 2.0 or later compiled with unix domain socket
+# support option.
+# Mosquitto must have proper privileges to handle the path. 
+# Mosquitto can create simultaniously both TCP and unix listener. 
+# listener unix:unix-path
 #listener
 
 # The maximum number of client connections to allow. This is 
@@ -842,3 +851,4 @@
 #max_log_entries
 #trace_level
 #trace_output
+
--- a/src/conf.c
+++ b/src/conf.c
@@ -12,6 +12,7 @@ and the Eclipse Distribution License is available at
  
 Contributors:
    Roger Light - initial implementation and documentation.
+   Tatsuzo Osawa - Add unix domain socket listener.
 */
 
 #include <config.h>
@@ -30,6 +31,7 @@ Contributors:
 #ifndef WIN32
 #  include <netdb.h>
 #  include <sys/socket.h>
+#  include <unistd.h>
 #else
 #  include <winsock2.h>
 #  include <ws2tcpip.h>
@@ -1132,17 +1134,43 @@ int config__read_file_core(struct mosquitto__config *config, bool reload, const 
 							return MOSQ_ERR_NOMEM;
 						}
 						tmp_int = atoi(token);
-						if(tmp_int < 1 || tmp_int > 65535){
-							log__printf(NULL, MOSQ_LOG_ERR, "Error: Invalid port value (%d).", tmp_int);
-							return MOSQ_ERR_INVAL;
+#ifndef WIN32 	// Listeners for unix domain socket
+						if(strlen(token) > 5 && tmp_int == 0){
+							token = strtok_r(token, ":", &saveptr);
+							if(token && strcmp(token, "unix")) {
+								log__printf(NULL, MOSQ_LOG_ERR, "Error: Invalid listener string.");
+								return MOSQ_ERR_INVAL;
+							}
+							cur_listener = &config->listeners[config->listener_count-1];
+							memset(cur_listener, 0, sizeof(struct mosquitto__listener));
+							cur_listener->use_unixsocket = true;
+						}else
+#endif
+						{
+							if(tmp_int < 1 || tmp_int > 65535){
+								log__printf(NULL, MOSQ_LOG_ERR, "Error: Invalid port value (%d).", tmp_int);
+								return MOSQ_ERR_INVAL;
+							}
+							cur_listener = &config->listeners[config->listener_count-1];
+							memset(cur_listener, 0, sizeof(struct mosquitto__listener));
 						}
-						cur_listener = &config->listeners[config->listener_count-1];
-						memset(cur_listener, 0, sizeof(struct mosquitto__listener));
 						cur_listener->protocol = mp_mqtt;
 						cur_listener->port = tmp_int;
 						token = strtok_r(NULL, "", &saveptr);
 						if(token){
 							cur_listener->host = mosquitto__strdup(token);
+#ifndef WIN32 	// Listeners for unix domain socket
+							if(cur_listener->use_unixsocket){
+								if(access(cur_listener->host, F_OK) == 0){
+									log__printf(NULL, MOSQ_LOG_ERR, "Error: The path of unix domain socket already exist.");
+									return MOSQ_ERR_INVAL;
+								}
+								if(errno != ENOENT){
+									log__printf(NULL, MOSQ_LOG_ERR, "Error: The path of unix domain socket is illegal.");
+									return MOSQ_ERR_INVAL;
+								}
+							}
+#endif
 						}else{
 							cur_listener->host = NULL;
 						}
@@ -1958,3 +1986,4 @@ static int conf__parse_string(char **token, const char *name, char **value, char
 	}
 	return MOSQ_ERR_SUCCESS;
 }
+
--- a/src/mosquitto.c
+++ b/src/mosquitto.c
@@ -12,6 +12,7 @@ and the Eclipse Distribution License is available at
  
 Contributors:
    Roger Light - initial implementation and documentation.
+   Tatsuzo Osawa - Add unix domain socket listener.
 */
 
 #include "config.h"
@@ -34,6 +35,7 @@ Contributors:
 
 #ifndef WIN32
 #  include <sys/time.h>
+#include <unistd.h>
 #endif
 
 #include <errno.h>
@@ -135,6 +137,27 @@ int drop_privileges(struct mosquitto__config *config, bool temporary)
 #endif
 	return MOSQ_ERR_SUCCESS;
 }
+
+#ifndef WIN32
+int drop_privileges_pathname(char *user, char *pathname)
+{
+	struct passwd *pwd;
+
+	if(geteuid() == 0){
+		if(user && strcmp(user, "root")){
+			pwd = getpwnam(user);
+			if(!pwd){
+				log__printf(NULL, MOSQ_LOG_ERR, "Error: Invalid user '%s'.", user);
+				return 1;
+			}
+			if(chown(pathname, pwd->pw_uid, pwd->pw_gid) == -1){
+				log__printf(NULL, MOSQ_LOG_WARNING, "Warning: Can't set proper privileges of unix domain socket path.");
+			}
+		}
+	}
+	return MOSQ_ERR_SUCCESS;
+}
+#endif
 
 int restore_privileges(void)
 {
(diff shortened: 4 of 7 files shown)


ralight:
Hi, thanks for your contribution. I like this in principle but have a few small points to discuss.

I don't really like the "port == 0 means unix domain socket" definition that's has been implicitly added. Given that the config file is `listener port-number` at the moment I'm not sure what the best thing to do here would be. It needs some thought though. Maybe it could be `listener port-number|unix` for the config and then internally add a new entry to the protocols enum (we can argue over whether a unix socket is a protocol or a transport, but it's consistent with the way it is being used so far). 

I would be grateful if you could add some documentation for this new feature to both man/mosquitto.conf.5.xml and mosquitto.conf - it doesn't need to be perfect.

Could you please add `{...}` around src/net.c:196 and src/conf.c:1137 until the end of the block?

There is also some inconsistent indentation in various places.

toast-uz:
Thank you very much for your comment. I will revise it.

toast-uz:
Hi Roger,
Adding a new entry to the protocols enum seems disable the possibility
of using Websockets on UINX sockets. Should I add a new entry to the
struct mosquitto__listener itself?
Regards,

ralight:
You're right of course. I'm thinking about the best way to do this.

toast-uz:
I've revised this considerably. 
Now the unix domain socket listener supports mqtt, mqtt-tls, and websockets. To use websockets, need libwebsockets 2.0 or later compiled with unix domain socket support option. See mosquitto.conf or mosquitto.conf(5) for more information.

ralight:
Do you have a definite need for unix domain sockets, or do you just think of this as a nice feature? Do other MQTT tools have unix socket support?

toast-uz:
My demand is described at the top of this conversation. I don't know other mqtt tools support UDS, but supporting UDS is a popular feature of a web server.

karlp:
Please also note, it's pretty much no longer the case that UDS offers a performance gain over TCP sockets.  It _was_, back in 2005 sort of time frame, but (unsuprisinginly) there's been massive work done on making TCP sockets faster, and it's simply no longer the case.  You're adding code and complexity for something that's no longer relevant.

karlp:
you've got more than 64k local clients?

toast-uz:
Not yet, but I must prepare.

ralight:
After some consideration, I've decided that the benefits of this addition are outweighed by the increased code complexity. That's not to say it couldn't be revisited in the future. Thanks for taking the time to try it out.

kiranpradeep:
@toast-uz I was running Mosquitto broker and client in different Docker containers with in a single machine. The TCP performance is very good when both runs on same machine (loopback) as native apps. But as dockerized applications, it is not as fast. It could be partially because Docker's slow internal TCP networking. I was looking at mapping a UDS into both containers and let the MQTT client and broker communicate over UDS. I felt this was a useful feature as I am looking to publish over 200k messages. Could this be reconsidered?

karlp:
@kiranpradeep or, you know, fix docker.  You already demonstrated that it wasn't a TCP problem as it was fine with both on the same machine.

karlp:
Bear in mind that the referenced redis tests document was testing with linux kernel 2.6.16.  This was definitely known at the time that UDS was faster than TCP localhost.  DEFINITELY remeasure if you really have concerns about this still.

karlp:
(it's heavily dependent on message sizes too, there's no magic buller of "just use UDS" otherwise everything would always be doing that)

andrewchambers:
@ralight unix domain sockets have security benefits because they support access controls on multi user systems.

karlp:
@andrewchambers you know you can use client certs right? no need to worry about ssh and user accounts?

andrewchambers:
@karlp yes, but the point is why someone might want unix domain sockets. Many people already have secure ssh infrastructure, its just annoying to add more things when it shouldn't be needed.

fpagliughi:
I have a little C++ socket library, _sockpp_, which has some basic echo client/server example apps. I just added test clients to measure the rate to echo messages (send then receive) using TCP v4 vs UDS.

This was by no means a scientific study, but I ran tests on my desktop PC and a couple of maker boards (BeagleBone, RPi), and in nearly every case UDS was twice as fast as TCP/localhost!!!

That's a significant performance increase,

These all generally had fairly recent Debian or Ubuntu distro's and were mostly kernel 4.14 or newer. @karlp can you site the sources claiming that TCP is now as fast as UDS. I'm not seeing that in the real world. Is there something that needs to be configured?

toast-uz:
Hi there, thank you for expecting my closed PR. Honestly the performance of my PR was as same as using TCP socket. Although it might be better under the current enviroment.

I expected Mosquitto as a component of a massive IoT platform. Therefore I made my effort for Mosquitto about the performance improvement and connectivity with the other component. In fact, my puppose for this PR was improving the connectivity with NGINX. Personally I like to use TLS features of NGINX better than Mosquitto. It was under the thought of the UNIX philosophy's "Worse is better" .

But as a result, the position has been owned by some mega cloud platforms. And now, Mosuqitto is expected as a small, compact, and completely packaged broker on the edge computers such as Raspberry PI. Therefore, I guess this PR has less priority.

fpagliughi:
> And now, Mosuqitto is expected as a small, compact, and completely packaged broker on the edge computers such as Raspberry PI. Therefore, I guess this PR has less priority.

Actually that's **exactly** why I think this is a priority. At least for one of my typical use cases.

I use MQTT in the standard way a lot of the time - where there are remote devices calling home with data. But quite often I use Mosquitto as an internal message bus in an embedded Linux gateway/controller. Usually it's done because I also bridge it and send some of the traffic to a remote server. So keeping all MQTT is a pretty nice setup.

The problem is that with a small board roughly the power of a BeagleBone, the trip from one client through the broker to the other via localhost can take 2-5ms, depending on the board, traffic, CPU load, etc. That's a pretty long time. Cutting that down by 50%, 30%, or even 20% would be quite helpful.

toast-uz:
@fpagliughi , I understood your situation. I guess the latency becomes better by using UDS, because it does not have TCP handshake overhead.

When I made this PR, I'd measured the performance only as packets per second.

ralight:
@fpagliughi The develop branch has got unix domain support. It is implemented differently to this pull request. The main reason that I changed my mind on this was the idea of having localhost only listener, where the access control was through the socket permissions. It's much easier to isolate a trusted listener that way.

fpagliughi:
Awesome! Yes. Very true. I hadn't thought to argue the security aspect. But that is big. 
I will look at adding support to a couple of the Paho libraries to start testing it.

dlitz:
Hm, I just got burned by this. 😕

I had mistakenly assumed that mosquitto would be able to connect on a unix socket, since it's a pretty standard thing I've come to expect from well-behaved Unix backend services.

Since people seem to be asking "why would you want this", I'll try to explain:

Yes, the main reasons for my wanting this are security and administrative simplicity.  Performance isn't a concern: my use-case is a centralized "hub" server that uses socket activation to reach out to a random collection of IoT devices, some of which are just consumer devices that I've rooted---Philips Hue, for example, and various wifi surveillance cameras (with dubious GPL-compliance, heh)---so I don't control the entire firmware image.  And the Hue device still does automatic software updates, so I prefer to touch their configuration as little as possible.

These devices often have dropbear pre-installed, and it's convenient to be able to just drop a small authorized_keys file onto them (and maybe an updated dropbear binary), and that one binary simultaneously takes care of remote administration, file transfers, and any secure socket connections into the device that I want to make from a remote host.

On the "hub"-side, though, this can very quickly become a mess of dozens of TCP listeners on arbitrarily-numbered ports that don't mean very much, and it's also difficult to do finer-grained access-control than "everything on localhost is trusted" (ugh).  With Unix sockets, I can put everything into a hierarchy under /run and control access via directory traversal permissions (and bind mounts, if I need something more wacky).  With unix sockets, plus systemd or ucspi socket activation, the work of setting up and tearing down connections is pretty much done for me automatically, and it's a _lot_ easier to read and avoid mistakes with something like "unix:/run/sshfwd/hue0-mqtt/sock" versus where everything looks like a subtle variant of "localhost:10002".

I had everything almost ready to go in pretty much just 3 files (plus the usual SSH key-management stuff that I do as a matter of course):

```
# cat /etc/systemd/system/sshfwd-hue0-mqtt.socket
[Unit]
Description=SSH port-forward to mosquitto running on Philips Hue bridge

[Socket]
ListenStream=/run/sshfwd/hue0-mqtt/sock
Accept=yes
SocketMode=0660
SocketGroup=mosquitto

[Install]
WantedBy=sockets.target
```
```
# cat /etc/systemd/system/sshfwd-hue0-mqtt@.service
[Unit]
Description=SSH port-forward to mosquitto running on Philips Hue bridge

[Service]
Type=oneshot
User=sshfwd
ExecStart=/usr/bin/ssh -F /etc/sshfwd/config -qW localhost:1883 hue0
StandardInput=socket
StandardError=journal
TimeoutStopSec=5
```
```
# cat /etc/sshfwd/config
IdentityFile /etc/sshfwd/id_rsa
IdentityFile /etc/sshfwd/id_ed25519
UserKnownHostsFile /etc/sshfwd/known_hosts
HashKnownHosts no
CheckHostIP no
BatchMode yes

Host hue0
    HostKeyAlias hue0.dlitz.net
    User root
    #HostName hue0.lan
    HostName 10.159.1.211
    KexAlgorithms +diffie-hellman-group14-sha1,diffie-hellman-group1-sha1
```

Abstracting connections like this also lets me do things like proxy connections through VPNs or even hardware serial ports.

I had hoped to be able to put something like `address /run/sshfwd/hue0/mosquitto` into my mosquitto.conf, but sadly, that doesn't work.

And just to highlight the _necessity_ of unix sockets on a modern machine, here's a partial list of the sockets on my laptop (installed by the distro):

```
$ sudo find /run -type s
/run/docker.sock
/run/wpa_supplicant/p2p-dev-wlp2s0
/run/wpa_supplicant/wlp2s0
/run/docker/metrics.sock
/run/docker/libnetwork/9582ba17845f.sock
/run/containerd/s/7265fecbda45c8a7f859f35abd49bad53ee16a71b22f20de288b0bb8eb38ce10
/run/containerd/containerd.sock
/run/containerd/containerd.sock.ttrpc
/run/vmware/usbarbitrator-socket
/run/uml-utilities/uml_switch.ctl
/run/irqbalance/irqbalance1625.sock
/run/uuidd/request
/run/snapd-snap.socket
/run/snapd.socket
/run/pcscd/pcscd.comm
/run/libvirt/virtlogd-admin-sock
/run/libvirt/virtlogd-sock
/run/libvirt/virtlockd-admin-sock
/run/libvirt/virtlockd-sock
/run/libvirt/libvirt-sock-ro
/run/libvirt/libvirt-admin-sock
/run/libvirt/libvirt-sock
/run/dbus/system_bus_socket
/run/cups/cups.sock
/run/avahi-daemon/socket
/run/acpid.socket
/run/user/1000/zim-0.73.4-e63203e4
/run/user/1000/kio_http_cache_cleaner
/run/user/1000/spectacleksUuoI.6.slave-socket
/run/user/1000/spectaclesTdwrL.13.slave-socket
/run/user/1000/spectacleSkGcrv.12.slave-socket
/run/user/1000/spectaclebvFIDl.9.slave-socket
/run/user/1000/spectacleJRVhvk.8.slave-socket
/run/user/1000/spectacleNssWIq.8.slave-socket
/run/user/1000/speech-dispatcher/speechd.sock
/run/user/1000/klauncherwIMWli.1.slave-socket
find: ‘/run/user/1000/doc’: Permission denied
/run/user/1000/kdeinit5__0
/run/user/1000/kwallet5.socket
/run/user/1000/keyring/control
/run/user/1000/mysqld@adleman.dlitz.net/mysqld.sock
/run/user/1000/snapd-session-agent.socket
/run/user/1000/pulse/native
/run/user/1000/pk-debconf-socket
/run/user/1000/pipewire-0
/run/user/1000/bus
/run/user/1000/gnupg/S.gpg-agent
/run/user/1000/gnupg/S.gpg-agent.ssh
/run/user/1000/gnupg/S.gpg-agent.extra
/run/user/1000/gnupg/S.gpg-agent.browser
/run/user/1000/gnupg/S.dirmngr
/run/user/1000/systemd/private
/run/user/1000/systemd/notify
/run/user/1000/systemd/inaccessible/sock
/run/samba/ncalrpc/DEFAULT
/run/samba/ncalrpc/np/winreg
/run/samba/ncalrpc/np/srvsvc
/run/samba/ncalrpc/np/netlogon
/run/samba/ncalrpc/np/lsarpc
/run/samba/ncalrpc/np/lsass
/run/samba/ncalrpc/np/samr
/run/samba/ncalrpc/np/netdfs
/run/samba/ncalrpc/np/wkssvc
/run/samba/ncalrpc/np/spoolss
/run/samba/ncalrpc/np/svcctl
/run/samba/ncalrpc/np/ntsvcs
/run/samba/ncalrpc/np/plugplay
/run/samba/ncalrpc/np/eventlog
/run/samba/ncalrpc/np/initshutdown
/run/samba/ncalrpc/np/mdssvc
/run/apt-cacher-ng/socket
/run/lvm/lvmpolld.socket
/run/systemd/io.system.ManagedOOM
/run/systemd/private
/run/systemd/resolve/io.systemd.Resolve
/run/systemd/fsck.progress
/run/systemd/journal/io.systemd.journal
/run/systemd/journal/socket
/run/systemd/journal/stdout
/run/systemd/journal/dev-log
/run/systemd/journal/syslog
/run/systemd/userdb/io.systemd.DynamicUser
/run/systemd/userdb/io.systemd.Machine
/run/systemd/notify
/run/systemd/inaccessible/sock
/run/udev/control
```

Imagine trying to allocate TCP port numbers for each one of those (and avoid clashes, and keep them straight in your head or even /etc/services, and write iptables rules for fine-grained access control...).  The developers of every individual service could probably make excuses and ask "do you really _need_ unix sockets here?", but the necessity comes from the having to deal with them all in the aggregate.

Another use said they decided to use redis rather than deal with this, and I entirely sympathize, because in any moderately complex setup, cramming everything through the tiny namespace of a 16-bit TCP port number is just---exhausting.

imo, backend services should never assume they're going to be able to send real packets of an actual network.  Sometimes, we're buried 2 or 3 layers deep in IPC, and services that can deal with (1) unix sockets, and (2) arbitrary file descriptors (as much as practical) are going to be a lot more versatile in those situations, and go a long way toward helping us keep our sanity.  The beauty of Unix is being able to take all these building blocks and knowing that we can assemble them in arbitrary ways as requirements emerge.  It can be frustrating when random components don't fit together as expected, and soul-sucking when when you google it, and upstream devs seem reluctant to accept patches because they don't see the point using Unix sockets on Unix...

(Apologies for bad grammar and long-winded comment.  Chalk it up to the pandemic & ADHD.)

fpagliughi:
Didn't mosquitto 2.0 add Unix socket support?

dlitz:
I don't see it as an option in the docs for outgoing bridge connections.

PR #484, opened 2017-07-08
Add Epoll.

toast-uz:
Added epoll for Linux.
Using epoll system-call instead of poll, mosquitto can get performance improvement especially in handling huge concurrent connections. I confirmed it on my test environment.
Compile option: "$ make binary WITH_EPOLL=yes" to try.

Signed-off-by: Tatsuzo Osawa (toast-uz) #


nanrenhaiyang:
I'm really grateful to you. How can i use it with cmake。

toast-uz:
You don't need to run cmake and can start "make clean" and "make binary
WITH_EPOLL=yes" directly on your LInux, although I have not yet arranged
CMakeLists.txt. After making, you can complete install by "make install
WITH_DOCS=no".

2017-07-13 15:27 GMT+09:00 nanrenhaiyang <notifications@github.com>:

ralight:
Thanks for this, it looks really good but does conflict with the changes for 1.4.14 that are now merged into the develop branch. Could you take a look? It should be just around the websockets changes.

Issue #327, opened 2016-12-03
mosquitto_loop crash on iOS at line FD_SET(mosq->sockpairR, &readfds);


haok3166:
Hi there,

I am using the latest mosquitto lib on iOS to receive streaming data from mqtt servers. It works perfectly except there are some crashes in mosquitto_loop method. However, I could not reproduce this crash on local tests. I saw the crash report from crashlytics which has around 0.5% crash rate.

			if(mosq->sockpairR != INVALID_SOCKET && FD_ISSET(mosq->sockpairR, &readfds)){
#ifndef WIN32
				if(read(mosq->sockpairR, &pairbuf, 1) == 0){
				}
#else
				recv(mosq->sockpairR, &pairbuf, 1, 0);
#endif
				/* Fake write possible, to stimulate output write even though
				 * we didn't ask for it, because at that point the publish or
				 * other command wasn't present. */
				FD_SET(mosq->sock, &writefds);
			}

The crash always happens on line 963 FD_SET(mosq->sock, &writefds);, I have tried to comment out this line and the app has no crash on our test iOS devices. I am wondering if there is a side effect doing this? Will this change cause crash once it is released to App Store?

I really apprecite your help and thank you for creating such great MQTT product.

Best Regards

ralight:
The sockpairR/W sockets are there to ensure if you call e.g. mosquitto_publish() when you're in threaded operation and the other thread is stuck in select(), that it can break out of select() before the timeout.

The call for FD_SET() is modifying the output of select() so that the subsequent code thinks that mosq->sock was actually writable - so not having to wait until the next call to select. There are two reasons I can think of that would cause this - either mosq->sock is too big - which it could be if you're opening lots of connections in one app (most likely >1024). More likely, if mosq->sock has been closed just before this call and set to -1, then FD_SET(-1, ...) would certainly crash.

The obvious solution is to check the validity of mosq->sock just before the call to FD_SET, so

    if(mosq->sock >= 0) FD_SET(...);

Would it be possible to test that to confirm the hypothesis?

Ultimately, commenting out the call to FD_SET() will at its worst just potentially introduce some latency into outgoing messages, so it may not be the end of the world for you.

haok3166:
Hi Roger,

Thank you very much for your reply.

[...]

Issue #63, opened 2016-03-15
unsafe use of fd_set


ralight:
migrated from Bugzilla [#464632](https://bugs.eclipse.org/bugs/show_bug.cgi?id=464632)
status ASSIGNED severity *critical* in component *Mosquitto* for *1.4*
Reported in version *1.4* on platform *PC*
Assigned to: Roger Light

On 2015-04-14 13:38:49 -0400, bd walker wrote:

> There are 2 issues I spotted with this:
> 1) maxfd should be of type mosq_sock_t
> 2) fd_set could be unsafe as FD_SETSIZE is often 1024 on Linux and Darwin platforms. If maxfd is >= 1024, there is a chance of overflowing fd_set.
> 
> 
> int mosquitto_loop(struct mosquitto *mosq, int timeout, int max_packets)
> {
> #ifdef HAVE_PSELECT
> 	struct timespec local_timeout;
> #else
> 	struct timeval local_timeout;
> #endif
> 	fd_set readfds, writefds;
> 	int fdcount;
> 	int rc;
> 	char pairbuf;
> 	int maxfd = 0;
> 
> -------------------------------
> 
> Indeed, I once encountered "Illegal access" on "mosq->sock" (see below), with mosq being a NULL, in iOS 8 x86_64 emulator.
> 
> 
> 
> #ifdef HAVE_PSELECT
> 	fdcount = pselect(maxfd+1, &readfds, &writefds, NULL, &local_timeout, NULL);
> #else
> 	fdcount = select(maxfd+1, &readfds, &writefds, NULL, &local_timeout);
> #endif
> 	if(fdcount == -1){
>                .....
> 	}else{
> 		if(mosq->sock != INVALID_SOCKET){
> 
> 
> If you look at this: http://eli.thegreenplace.net/2011/09/06/stack-frame-layout-on-x86-64/ ,
> the first argument "mosq" lives in the %rdi register. However, since this register is not preserved across function calls, it is saved on the stack:
> 
> Start of mosquitto_loop:
> 0x10e6dcc00 <+0>:    pushq  %rbp
>     0x10e6dcc01 <+1>:    movq   %rsp, %rbp
>     0x10e6dcc04 <+4>:    subq   $0x180, %rsp
>     0x10e6dcc0b <+11>:   movq   %rdi, -0x10(%rbp)  <---- save %rdi into stack
>     0x10e6dcc0f <+15>:   movl   %esi, -0x14(%rbp)
[...]

PR #486, opened 2017-07-11
Fix iOS crash issues

Zard1096:
Relate to issues #327 and #63.
mosq->sock may be closed before FD_SET(mosq->sock, &writefds) and
FD_ISSET(mosq->sock, &writefds) but after judgement in line 947
if(mosq->sock != INVALID_SOCKET). FD_SET(-1, ...) and FD_ISSET(-1, ...)
would certainly crash.

Signed-off-by: Zard1096 <mr.zardqi@gmail.com>


ralight:
That's great, thanks very much. I've committed the change from the patch rather than by this PR directly.

PR #493, opened 2017-07-16
Replace getdtablesize() with sysconf(_SC_OPEN_MAX)

fornwall:
From http://man7.org/linux/man-pages/man3/getdtablesize.3.html:

"It is not specified in POSIX.1; portable applications should employ
sysconf(_SC_OPEN_MAX) instead of this call."

Specifically this fixes a build failure on Android which does not
have getdtablesize().


ralight:
That sounds reasonable, thanks very much.

PR #498, opened 2017-07-18
Fix the TLS handshake problem if PSK has leading zero

askawu:
Incorrect psk will be provided by psk_server_callback() because leading zero
is skipped by BN_bn2bin() and BN_num_bytes()

Signed-off-by: Aska.Wu <askawu@gmail.com>

### DIFF
--- a/lib/util_mosq.c
+++ b/lib/util_mosq.c
@@ -327,19 +327,33 @@ int _mosquitto_hex2bin(const char *hex, unsigned char *bin, int bin_max_len)
 {
 	BIGNUM *bn = NULL;
 	int len;
+	int leading_zero = 0;
+	int start = 0;
+	int i = 0;
+
+	/* Count the number of leading zero */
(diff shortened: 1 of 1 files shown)


ralight:
Thanks very much, this is a good catch. As long as you were using mosquitto+libmosquitto then nothing would appear to be incorrect because of the same code in both cases. It's good to have those assumptions tested.

I've applied this via [d40d777](https://github.com/eclipse/mosquitto/commit/d40d7772d37f1de587dfd9bc49d374d8d5b393b7).

Thanks for your contribution!

imZack:
@askawu which version of openssl do you use?
I have applied this patch into my code and it doesn't work but the un-patched one does.

ralight:
Right, so in the first case the leading zeros were trimmed which is incorrect.

imZack:
Ah, got it. I think i missed to patch the "clients". Thanks @ralight

PR #500, opened 2017-07-20
support for 1.4.14 into docker

agrosjea:
As I encounter issue with 1.4.12, I needed to test with 1.4.14

Here is a pull request based on alpine:edge

please note that for reason that I ignore, docker-entrypoint.sh didnt wanted to run before I added specific permission, which isnt in your version.

### DIFF
--- a/docker/1.4.14/Dockerfile
+++ b/docker/1.4.14/Dockerfile
@@ -1,0 +1,13 @@
+FROM alpine:edge
+MAINTAINER David Audet <david.audet@ca.com>
+
+LABEL Description="Eclipse Mosquitto MQTT Broker"
+
+RUN apk --no-cache add mosquitto=1.4.14-r0 && \
+    mkdir -p /mosquitto/config /mosquitto/data /mosquitto/log && \
+    cp /etc/mosquitto/mosquitto.conf /mosquitto/config && \
+    chown -R mosquitto:mosquitto /mosquitto
+
+COPY docker-entrypoint.sh /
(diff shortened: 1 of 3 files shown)


ralight:
Thanks for doing this. I have some suggestions before I will accept the PR:

* The permissions issue is down to the new docker-entrypoint.sh file not having the 755 permissions when added to the repository - please fix that and remove the chmod line.
* Please squash all of these commits into a single commit.
* You should ensure that the signed-off footer on the commit message is correct (use "git commit -s")
* You also need to sign the ECA: http://www.eclipse.org/legal/ECA.php

Once you've done all that, you can force push (git push -f) to your existing branch and it will update this PR.

agrosjea:
Dear rogers, 

Thanks for doing such a great review of such a poor contribution 😃 
I made my best to correct everything but still see that ip-validation didnt passed. do I need to do something on my side ?

ralight:
This is very nearly ready for merging, thanks. You just need to change one tiny thing to make the commit valid - there should be a blank line between the actual commit message and the footer (the Signed-off-by line). I don't make the rules, this is just how it has to be. If you blank out the first signed-off-by line (the .local one) it should do the job. The second email address is fine, I've checked the ECA.

agrosjea:
I'm on it !

agrosjea:
ok the ip-validation has passed now, thanks for your time and advices.

ralight:
Great! Thanks for your patience and your contribution.

PR #501, opened 2017-07-20
Add test for persistent subscription

PierreF:
The PR add test for persistent subscription.
This test show a bug currently present in develop branch (and not in master).

The issue is that subscription are not correctly persisted in the database. Without the test, we can reproduce the bug with:
```
$ cat mosquitto.conf 
persistence true
listener 1888
$ rm -f mosquitto.db
$ ./mosquitto -c mosquitto.conf&
$ ./client/mosquitto_sub -p 1888 -q 1 -t thetopic -c -i test   # Stop it as soon as it connected to broker
$ ./client/mosquitto_pub -p 1888 -q 1 -t thetopic -m "message sent before restart"
# Stop the broker and restart it
$ ./mosquitto -c mosquitto.conf&
$ ./client/mosquitto_pub -p 1888 -q 1 -t thetopic -m "message sent after restart (not received)"
$ ./client/mosquitto_pub -p 1888 -q 1 -t " /thetopic" -m "message sent after restart on another topic"
$ ./client/mosquitto_sub -p 1888 -q 1 -t thetopic -c -i test
message sent before restart
message sent after restart on another topic
```

By reading mosquitto.db, the topic stored in mosquitto.db is not "thetopic", but "<whitespace>/thetopic".

This issue was introduced by 883af8a, which prefix all topics by a whitespace (except $SYS). I'm no sure to understand why that whitespace is added and tried to remove [the if that add it](https://github.com/eclipse/mosquitto/blob/c26b852c530d471e8a51439f78b4d0821bfc790d/src/subs.c#L186), while it worked for basic subscription (and presitence), it breaks 03-pattern-matching test.

### DIFF
--- a/test/broker/11-persistent-subscription.conf
+++ b/test/broker/11-persistent-subscription.conf
@@ -1,0 +1,3 @@
+port 1888
+persistence true
+persistence_file mosquitto-test.db
--- a/test/broker/11-persistent-subscription.py
+++ b/test/broker/11-persistent-subscription.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python
+
+# Test whether a client subscribed to a topic receives its own message sent to that topic.
+
+import inspect, os, sys
+# From http://stackoverflow.com/questions/279237/python-import-a-module-from-a-folder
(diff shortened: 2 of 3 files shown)


ralight:
I think the problem is the test at persist.c:283.

The space is there to help segregate the $SYS tree (or other $ topics) from everything else. There might be a better way.

Issue #497, opened 2017-07-17
SSL with Socket


achest:
I would like to use secure websocket connection. Is this implemented? Have you  a sample configuration for me? Which test client can I use?

My Configuration: 
Mosquitto 1.4.11, OS:  Fedora / Ubuntu (inside Docker).
Client. NodeJS MQTT

It works without encryption or at normal port. With SSL encryption not.
```
user mosquitto
port 8883
tls_version tlsv1.2
cafile /mqtt/config/m/ca
certfile /mqtt/config/m/crt
keyfile /mqtt/config/m/key
require_certificate true
log_type all
connection_messages true
websockets_log_level 31
listener 1443
protocol websockets
tls_version tlsv1.2
cafile /mqtt/config/m/ca
certfile /mqtt/config/m/crt
keyfile /mqtt/config/m/key
```

Debug says SSL Context is null.  Many thanks,

ralight:
That looks fine. Could you try running `openssl s_client -connect <host>:1443 -showcerts` and pasting the result?

achest:
It's looks fine. 

Which socket client you are using for testing ?

```
[achest@pc33 test-server]$ openssl s_client -connect localhost:1443 -showcerts
CONNECTED(00000003)
depth=1 CN = test-connect, C = de
verify error:num=19:self signed certificate in certificate chain
---
Certificate chain
 0 s:/CN=ess/O=test-connect/C=de
   i:/CN=test-connect/C=de
-----BEGIN CERTIFICATE-----
MIIDADCCAeigAwIBAgIJAOIk6lnGm5dkMA0GCSqGSIb3DQEBDQUAMCYxFzAVBgNV
BAMMDmV4Y2VldC1jb25uZWN0MQswCQYDVQQGEwJkZTAeFw0xNzAzMjIxNzE5MTNa
Fw0zMjAzMTgxNzE5MTNaMDQxDDAKBgNVBAMMA2VzczEXMBUGA1UECgwOZXhjZWV0
LWNvbm5lY3QxCzAJBgNVBAYTAmRlMIIBIjANBgkqhkiG9w0BAQEFAAOCAQ8AMIIB
CgKCAQEAnJpbFUASC3UJxYjR6kLd3TM1T40lH+a2d+rNZK68Bplh1l0xc8r4tBVh
nnytCy0t9Yqxoz3yUPL5jAR+ih7nIYNr0PnSH5lzojrCjBa8SnVCa8mmYLDKufyL
[...]

PR #504, opened 2017-07-26
Fix a bug for use_identity_as_username for websockets

achest:
I have already opened an issue for this: #497

If parameter use_identity as_username is true, that works only for a MQTTS connection. If websockets are used, the username is not filled with content from UserCertificate.

I found the place in websocket.c and has fixed the bug. Implementation of use_identity as_username is now uniform.
We also liked to use DN as username. Configurable as parameters in the listener section of the configuration file.

We would like to know if such patches are also welcome.
Many Thanks.
Best regards, Alexander
Signed-off-by: Alexander Chestnov <Alexander.Chestnov@exceet.de>

### DIFF
--- a/src/websockets.c
+++ b/src/websockets.c
@@ -158,6 +158,12 @@ static int callback_mqtt(struct libwebsocket_context *context,
 	int rc;
 	uint8_t byte;
 
+#ifdef WITH_TLS
+	X509 *client_cert = NULL;
+	X509_NAME *name;
+	X509_NAME_ENTRY *name_entry;
+#endif
+
 	db = &int_db;
 
 	switch (reason) {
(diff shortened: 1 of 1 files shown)


ralight:
Sorry for the delay looking at this. What you've done looks like it should work for you, but it's not quite right for the project.

I've spent some time on it this evening and believe I've fixed it with https://github.com/eclipse/mosquitto/commit/7943072b1f3bf68e820655408ec4ecb73fd8a3d9 , which is the latest commit on the fixes branch. Are you able to give it a test?

ralight:
I'm closing this with the assumption that it is fixed based on the changes described above. If the changes don't work for you, please open a new bug report.

gautaz:
@ralight Hello, I'm also in need of this feature.

Will it be included in a release any time soon ?

Issue #507, opened 2017-07-29
Test 08-ssl-bridge.py failed even on master branch.


toast-uz:
As I reported in mosquitto-dev ML before, test 08-ssl-bridge.py failed even on master branch. Seemed to fail to accept bridge on port 1883. I've tested on CentOS7.

```
$ make test
(snip)
./08-ssl-bridge.py
Warning: Bridge bridge_test using insecure mode.
1501332759: mosquitto version 1.4.14 (build date 2017-07-29 12:50:18+0000) starting
1501332759: Config loaded from 08-ssl-bridge.conf.
1501332759: Opening ipv4 listen socket on port 1889.
1501332759: Opening ipv6 listen socket on port 1889.
1501332759: Bridge local.ip-172-31-17-55.ap-northeast-1.compute.internal.bridge_test doing local SUBSCRIBE on topic bridge/#
1501332759: Connecting bridge bridge_test (localhost:1888)
1501332759: Warning: Unable to connect to bridge bridge_test.
1501332760: New connection from ::1 on port 1889.
1501332760: Socket error on client <unknown>, disconnecting.
1501332780: mosquitto version 1.4.14 terminating

Traceback (most recent call last):
  File "./08-ssl-bridge.py", line 37, in <module>
    (bridge, address) = ssock.accept()
  File "/usr/lib64/python2.7/ssl.py", line 881, in accept
    newsock, addr = socket.accept(self)
  File "/usr/lib64/python2.7/socket.py", line 202, in accept
    sock, addr = self._sock.accept()
socket.timeout: timed out
make[2]: *** [08] Error 1
make[2]: Leaving directory `/home/centos/mosquitto/test/broker'
make[1]: *** [test] Error 2
make[1]: Leaving directory `/home/centos/mosquitto/test'
make: *** [test] Error 2
```

ralight:
Looks like this is down to the async dns changes, it works fine without.

ralight:
I think the latest commit to `fixes` does the job, could you please confirm it works for you?

toast-uz:
Tried but unfortunately the result seemed same.

```
$ git log --oneline
a480a9d Fix 08-ssl-bridge.py test when using async dns lookups.
124ee1a [490] Further fix for auth related crashes.
(snip)
$ make test
(snip)
Warning: Bridge bridge_test using insecure mode.
1501415619: mosquitto version 1.4.14 (build date 2017-07-30 11:50:55+0000) starting
[...]

PR #508, opened 2017-08-04
Fix 08-ssl-bridge.conf to dial localhost by ipv4.

toast-uz:
See #507


ralight:
Thanks, a nice easy fix to a non obvious problem!

PR #516, opened 2017-08-13
MQTT version 5 (Experimental support).

toast-uz:
I added MQTT version 5 experimental support based on OASIS-MQTT working draft 15.

I've implemented the broker handling and sending MQTT v5 packets, and converting packets between connections v3.x and v5. In addition, mosquitto_pub & mosquitto_sub could talk v5 by option -V mqttv5. (Unfortunately, there are few clients talking v5 so far, I must make them.)

Sorry not enough supported even if mandatory features, and a lot of revised from original branch. I would like to get early feedbacks by many experts about my implementation policy heading for completed support. Of course, adding features will be grateful also.


toast-uz:
I checked the latest OASIS MQTT version 5 committee specification draft v1 starting the invitation to comment last week. There seemed no part influencing my PR.

See [Invitation to comment on MQTT V5.0](https://www.oasis-open.org/news/announcements/invitation-to-comment-on-mqtt-v5-0-ends-sept-8th)

ralight:
I've taken a quick look at this and have some early thoughts. First off, good work!

I would very much prefer to build this up commit by commit rather than with one huge commit. I realise that might be a pain for you now, but it makes it much easier to review. Try to make it a single topic per commit.

For example, adding the property support is a bit element of what you've done which hides some of the other work a bit. A first go at getting v5 support working would in my mind be to add awareness that properties exist, so that packets containing them can be parsed, but without parsing the actual properties. Once that base was working, a later commit could add proper parsing, and then further commits would each add support for the different property behaviours. Likewise separating out the client / library / broker components into separate commits should make things easier. I'm not particularly worried about lots of commits.

On a more specific note (again, based off just a quick look):

* You are modifying the persistence format, which is fine, but you must update the db version number and retain the ability to load the current version (only writing out to the new format is fine)
* On a similar note, you'll also need to update db_dump for the new format
* You've got some big macros, any reason why they need to be macros not functions?
* I'm not entirely sure I like the way you are approaching the properties, but it is a bit difficult to tell in amongst everything else that is going on and I've not thought about it much myself anyway so this may be the best approach.

ralight:
Another very important comment is that there need to be tests. Integrating the tests currently being written at https://github.com/eclipse/paho.mqtt.testing/tree/mqttv5 would be one option.

toast-uz:
Thank you for your comment.
I understood and will create new PR.
Having tried a lot of implementation will be helpful for promoting detailed PRs.

Should I close this PR ?

PR #519, opened 2017-08-17
MQTT version 5 (limited to basic packet transfer).

toast-uz:
Reimplemented MQTT version 5 support.
- Focus on the broker's features of basic packet transfer.
- Integrate test using paho.mqtt.testing
  (Need to revise 08-ssl-bridge.conf and skip 11-persistent-subscription.py for reaching the tests of 12-mqttv5.py)

Please evaluate.

### DIFF
--- a/lib/handle_pubackcomp.c
+++ b/lib/handle_pubackcomp.c
@@ -12,6 +12,7 @@ and the Eclipse Distribution License is available at
 
 Contributors:
    Roger Light - initial implementation and documentation.
+   Tatsuzo Osawa - Add mqtt version 5.
 */
 
 #include <assert.h>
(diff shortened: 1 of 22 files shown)


toast-uz:
I'd like to consult.

In this PR, rewrote various parts of "return 1" to "return MOSQ_ERR_PROTOCOL".
These are for catching error at src/read_handle.c, because MQTT v5 requests detailed DISCONNECT reason and I wanted to simplify the codes.  Some of "return 1" were remained, because I could not determine the meaning.

If current "return 1" means definitely "return MOSQ_ERR_NOMEM", these should be reseted.
If current "return 1" means disconnecting connection, these seemed correct. 

In other words, all "return a number" might be rewrote to some of MOSQ_ERRs ?

toast-uz:
Fixed UNSUBACK format did not correspond with v5.

Should I keep still waiting to merge?
I would like to add new commit because the journey for completing mqttv5 is very long and this PR was just a first step.

ralight:
Thank you for making this PR, but I'm going to close it because there is now mqtt 5 support (in the `mqtt5` branch) that covers what is here. I apologise for not using your work, the way I wanted to implement things would have meant I would have had to undo a lot of what you had already done, so it seemed better to start from what I already had - which is some work I started a long while back anyway.

PR #525, opened 2017-08-23
Support using SSL connection over SOCKS5 PROXY

binli71:
This patch adds support to use SSL over SOCKS5 PROXY.
The issue was:
- when SSL was used, all SOCKS5 messages would be encrypted as well
- SOCKS5 proxy server can't interpret the encrypted SOCKS5 message
- the connection would be rejected in this case.
The solution is to enable SSL configuration after SOCKS5 connection is
completed if proxy is used.

Signed-off-by: Bin Li <bin.li@windriver.com>

### DIFF
--- a/lib/net_mosq.c
+++ b/lib/net_mosq.c
@@ -653,7 +653,10 @@ int _mosquitto_socket_connect(struct mosquitto *mosq, const char *host, uint16_t
 	if(rc > 0) return rc;
 
 	mosq->sock = sock;
-	rc = _mosquitto_socket_connect_step3(mosq, host, port, bind_address, blocking);
+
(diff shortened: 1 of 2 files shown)


binli71:
Following test case have been done and all passed.
Port 1883:  MQTT
Port 8883:  MQTT (TLSv12).

UsingPort   Socks5-Proxy
----------------------------------
1883     yes    :  Passed
8883     yes    :  Passed
1883     no     :  Passed
8883     no     :  Passed

ralight:
This has been fixed in a separate PR, apologies I should have got to yours first. Thanks for your contribution anyway!

PR #532, opened 2017-08-27
Fix persistent subscription issue #501.

toast-uz:
Fixed persistent subscription issue #501 .

Struggling with this issue, I guess the magic prefix  of a whitespace might be kept, because a lot of subhier functions need a sort of root topics. Therefore, patched in persist.c directly for handling the whitespace specially.

I can add some codes to invalidate old version of persistent database.


PierreF:
Thanks, it indeed fix issue #501 :)

About database invalidation, I would say that nothing is needed. With your patch database from fixes branch works correctly.
I'm not sure migration from one commit in develop to another commit in develop is something we need to support.

toast-uz:
Closed this PR because I like #544 better than this.

PR #535, opened 2017-08-29
Bridge dynamic by a publish on $SYS topic.

Tifaifai:
Hi,

In this PR, we submit a new functionality for the bridge.
Our commit propose to manage bridge dynamically by a publish message on $SYS topic.
This new functionality permit to create and delete bridges after the config file step.

Our use case is IoT management with different brokers on different Embedded Systems.

Please evaluate.


toast-uz:
Currently $SYS provide broker statuses for reading only. Therefore, be careful to change it updatable for security reason. Any particular reason to make an external tool to change mosquitto.conf and in need make a separated PR only for reconfiguration bridge features dynamically.

Tifaifai:
No documentation / standard specifies the obligation to be read-only for $SYS.
We just propose to make it possible to read and write a sub topic hierarchy.
``` src/bridge_dynamic.c : $SYS/broker/bridge/new and $SYS/broker/bridge/del```

We underline the security risk on these two topics but we propos to use a acl_config.
``` man/mosquitto.8.xml : 'ACL resctiction file for securise bridge dynamic management' ```
Maybe put this config BY DEFAULT in the mosquitto.conf :
```
# ACL
acl_file /home/mosquitto/etc/mosquitto/mosquitto.acl
password_file /home/mosquitto/etc/mosquitto/mosquitto.passwd
allow_anonymous true
``` 
and in mosquitto.acl file : 
```
# This affects access control for clients with no username.
topic read $SYS/#

# This only affects clients with username "roger".
user roger
topic readwrite $SYS/broker/bridge/new
topic readwrite $SYS/broker/bridge/del

# This affects all clients.
pattern write $SYS/broker/connection/%c/state
```

The idea isn't to modify in mosquitto.conf but permit to create and modify different bridge after starting the mosquitto server.

Our use case is IoT context and we need to link with bridge many broker.

For example : 

BrokerA <-> BrokerB <-> BrokerC <-> BrokerD <-> BrokerE
If BrokerC break down (network / power supply ... problems for example) then : 
A supervisor application can re-link BrokerB with BrokerD without change mosquitto.conf and restarting each mosquitto server. This gives time to reintegrate the BrokerC if necessary.

The ```client/mosquitto_bridge``` is an example application like ```client/mosquitto_pub``` and ```client/mosquitto_sub``` ;-)

Our feature allows us to develop the management aspects of the bridge currently in its infancy.

toast-uz:
You are right the standard allows using $SYS to control broker. But I don't like adding it to mosquitto, because MQTT ID/Password security is not strong.

Could you use current bridge round_robin feature to maintain backup bridge connections? If not, I'd like to recommend to expand it to solve your situation more elegantly.

Tifaifai:
Effectively, in our example, the bridge round_robin feature permit to reconnect with an other broker which is already known. 
This feature permit to connect a new broker with default configuration file.
The supervisor application can discover a new broker and create the new bridge with all rest of system.

This feature work with different dynamic topologies network.

toast-uz:
So, I'd like to recommend you to make an external tool to change mosquitto.conf and improve bridge configurations can be reloaded. Increasing reloadable configurations might be a good enhancement. Of course, it's a just my personal comment.

toast-uz:
P.S.
I believe the PR of reloadable bridge configurations also become the first and appropriate step of your original thought.

Tifaifai:
It's not possible to reload mosquitto config file to remote broker with SIGHUP signal.
Because on remote server, we have only access to 1883 port, so just publish or subscibe.

A network mesh should be modify bridge topology without system acess.

toast-uz:
The situation seemed so special ! Basically I guess the implementation on mainline should make benefit for a lot of users. So, first of all, you must insist your requirement is generic.

Could you make a sort of 'agent' software in your embedded devices to connect with a management server to check the new bridge information periodically? It is not a server, no need to listen a port, so might satisfy your requirement.

Or, you can make your own private mosquitto. I think mosquitto's BSD license allows it.

toast-uz:
FYI, as one of the client-server type management architecture, OMA lwm2m might help you.
https://en.wikipedia.org/wiki/OMA_LWM2M

Tifaifai:
On our master base, branch bridge-dynamic-master, we have integrate also this functionality of bridge dynamique : [https://github.com/Tifaifai/mosquitto/commit/540a4f17d151bdba83ee7d94b6963640171da9ec](https://github.com/Tifaifai/mosquitto/commit/540a4f17d151bdba83ee7d94b6963640171da9ec)
[https://github.com/Tifaifai/mosquitto/commit/39b4bb3aad6b7f60f2de800e48c34c9ba30abf8a](https://github.com/Tifaifai/mosquitto/commit/39b4bb3aad6b7f60f2de800e48c34c9ba30abf8a)
[https://github.com/Tifaifai/mosquitto/commit/9248bd8673292006a30cfd8731b20043e1581d1a](https://github.com/Tifaifai/mosquitto/commit/9248bd8673292006a30cfd8731b20043e1581d1a)

Our ```A supervisor application can re-link...``` is identic of your ```a sort of 'agent' software in your embedded devices``` ;)

We work on Multi-Agent System (Artificial Intelligent). And this functionality permit to rest the same philosophy of Publish/Subscribe mecanism to dynamically scale a cluster of Mosquitto servers without re-starting jobs.

We believe that it is beneficial for the whole community to integrate into the Mosquitto project. Perhaps in the future it may be possible to improve this functionality.

karlp:
fwiw, as a third party I _definitely_ see the usecase and motivation for being able to add bridges like this.

One thing that maybe worth considering though, is that bridges can be in both directions.  If you want to add a bridge from A to B, you don't have to configure A to bridge out to B.  You can also add a bridge in on B from A.  (Sometimes mountpoints and ACLs can make this harder than you want, but bridges can go in either direction....)

toast-uz:
To summarize my thoughts:
a) Better to divide the requirements into:
    1) Add reloadable bridge config.
    2) Add $SYS updatable to control broker.
b) Should design carefully on 2) because of security, etc.
c) There may be some workarounds of 2),

Tifaifai:
Glad to see that this philosophy can please.

Indeed, we believe that the live modification of some server elements can be an important step forward for Mosquitto.

The MQTT standard for topic $SYS is still rather unclear and Mosquitto seems to be driving the process. We believe that evolving the standard in this direction can bring many solutions and better scalability.

It is obvious that some security problems may appear but our contribution processes messages in the topic $SYS with the same way as usual publishes and therefore benefits from the same security features as usual exchanges.

toast-uz:
If you want to proceed with (2), could you add a switch such as 'sys_allow_update [true | false]' in mosquitto.conf and set the default to false? Then, every users can keep security as same as before.

Tifaifai:
For you my friend Tatsuzo ;-)

Tifaifai:
And master branch commit ;-)

toast-uz:
Thanks !

Tifaifai:
And fixes branch commit ;-)
3 operational branches for the upgrading of mosquitto server !
Update soon ? @ralight ;-)

Tifaifai:
reopen with Tifaifai:bridge-dynamic-develop

cjmaynard:
I was asked to leave a comment here and a link back to issue #539 and then close that issue to track everything here, so this comment serves that purpose.

PR #531, opened 2017-08-25
Fix subs memory issue.

toast-uz:
Sorry for #528  closed accidentally.

------------
Rearranged subhier relation architecture and fixed memory issue #505 .

FYI, a sample dump of subhier after fix:

```
lv: children generation level
h: head of hash
f: following item of hash
p: parent of item
c: children of item

1503576769: lv:0 h:0x104f400   -> f:0x104f400   (p:(nil) c:(nil))
1503576769: lv:0 h:0x104f400   -> f:0x104f6d0 $SYS (p:(nil) c:0x104f860 $SYS)
1503576769: lv:1 h:0x104f860 $SYS -> f:0x104f860 $SYS (p:0x104f6d0 c:0x104fb30 broker)
1503576769: lv:2 h:0x104fb30 broker -> f:0x104fb30 broker (p:0x104f860 c:0x104fe00 version)
1503576769: lv:3 h:0x104fe00 version -> f:0x104fe00 version (p:0x104fb30 c:(nil))
1503576769: lv:3 h:0x104fe00 version -> f:0x10501d0 timestamp (p:0x104fb30 c:(nil))
1503576769: lv:3 h:0x104fe00 version -> f:0x10503f0 uptime (p:0x104fb30 c:(nil))
1503576769: lv:3 h:0x104fe00 version -> f:0x1050530 clients (p:0x104fb30 c:0x10505a0 total)
1503576769: lv:4 h:0x10505a0 total -> f:0x10505a0 total (p:0x1050530 c:(nil))
1503576769: lv:4 h:0x10505a0 total -> f:0x1050910 inactive (p:0x1050530 c:(nil))
1503576769: lv:4 h:0x10505a0 total -> f:0x1050a50 disconnected (p:0x1050530 c:(nil))
1503576769: lv:4 h:0x10505a0 total -> f:0x1050b50 active (p:0x1050530 c:(nil))
1503576769: lv:4 h:0x10505a0 total -> f:0x1050cb0 connected (p:0x1050530 c:(nil))
Comparing with before fix in #505 helps your understanding.
At last, the parent pointed the real parent.
```

Comparing with before fix in #505 helps your understanding.
At last, the parent pointed the real parent.

Signed-off-by: Tatsuzo Osawa (toast-uz)

PR #544, opened 2017-09-08
Fix for persistent subscription issue #501.

PierreF:
This is an alternative fix to #532.

While looking at subs.c, I've found out that always search twice in the subs hierarchy using the very first token.
For example with sub_add/sub_add_recurse:
* sub__add will use the first token (" " or "$SYS") to find the subhier [link](https://github.com/eclipse/mosquitto/blob/d3239920d789a16cb9c54c598f4fc7d5ef03f2e9/src/subs.c#L461) 
* sub__add will call sub__add_recurse with the same token (e.g. not token->next)
* sub__add_recurse will once more use the first token to find the subhier [link](https://github.com/eclipse/mosquitto/blob/d3239920d789a16cb9c54c598f4fc7d5ef03f2e9/src/subs.c#L312)

This could be also seen with the sub__tree_print (signal USR2):
Without the fix:
```
1504888920: Socket error on client test, disconnecting.
     
       
        thetopic (test, 1)
    $SYS
      $SYS
        broker
```

With the fix:
```
1504888794: Socket error on client test, disconnecting.
     
      thetopic (test, 1)
    $SYS
      broker
```

Note that $SYS is present twice, also note that "thetopic" is indented on more level and an additional line (for the " " root prefix) is also present.

### DIFF
--- a/src/subs.c
+++ b/src/subs.c
@@ -468,7 +468,7 @@ int sub__add(struct mosquitto_db *db, struct mosquitto *context, const char *sub
 		}
 
 	}
-	rc = sub__add_recurse(db, context, qos, subhier, tokens);
+	rc = sub__add_recurse(db, context, qos, subhier, tokens->next);
 
 	sub__topic_tokens_free(tokens);
 
@@ -490,7 +490,7 @@ int sub__remove(struct mosquitto_db *db, struct mosquitto *context, const char *
 
 	HASH_FIND(hh, root, UHPA_ACCESS_TOPIC(tokens), tokens->topic_len, subhier);
 	if(subhier){
-		rc = sub__remove_recurse(db, context, subhier, tokens);
+		rc = sub__remove_recurse(db, context, subhier, tokens->next);
 	}else{
 		printf("nope\n");
 	}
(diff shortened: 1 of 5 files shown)


toast-uz:
Great work!
I like this better than my PR #532 .

toast-uz:
Confirmed this PR worked properly even after my PR #531 (Fix subs memory issue).

ralight:
This needs looking at very carefully, I'm not sure the proposed solution is correct.

The reason for the duplicate "$SYS" and the empty "" is to enforce a separation between the "normal" tree and $SYS, so that subscribing to '#' does not subscribe to $SYS as well. This does need looking at more generally than just for $SYS.

Normally we wouldn't want to persist anything on $SYS, the only reason to do so is if there is a queued message for a user who has subscribed to it.

PierreF:
There is indeed an issue during persistence :/
I've added 2 tests:
* One to test that pattern subscription like "#" does NOT match $SYS (it works)
* One to test that such pattern are persisted... which show that "$SYS/broker/uptime" is not persisted.

I'll look to this that.

ralight:
There's definitely something not quite right with this, I'm having some inconsistencies when trying to load a persistence file that was created without this patch.

ralight:
This was closed by accident, but I think I'm going to leave it closed unless there is an update to move it on.

PR #545, opened 2017-09-08
Remove some test that was always true

PierreF:
This PR remove few code that is currently dead-code.

All assert(subhier) always pass, because for each case, we search the first token (" " or "$SYS") in db->subs which is initialized with this two entries.
And those root entires are never remove, the one removing branch is sub__remove_recurse but only operate on second level (first level is done by sub__remove)

For sub__tree_print, it's called with level=0 or level=level+1. So level is always > -1.


ralight:
This was merged on the command line due to the conflict, thanks for working on this.

https://github.com/eclipse/mosquitto/commit/6fc7cadb63477a438f56745872a2046ed7026fc4

PR #552, opened 2017-09-14
Use getopt_long() for getting args.

xiangjiangnan:


karlp:
What's the gain here? It already works, and getopt_long has lots more portability problems

ralight:
Thanks for your contribution, but as Karl says this doesn't really provide a benefit and introduces portability problems so I'm going to close it.

PR #562, opened 2017-09-25
Add support for auth-plugin per listener

### LINKED ISSUE #242: Configure auth plug-in per listener
It appears that auth plug-in setting is for all listeners. If so, is it possible to define auth plug-in per listener? 

We would like to authenticate the publisher running in-house using the client SSL cert, but consumers running on mobile devices using the security tokens. We are writing the plug-in for token validation, but are concerned that it would overwrite authentication by client cert, and setting the user id to cert's CN.

PierreF:
This PR includes first step to fix #242. It adds per-listener authentication/authorization settings.

The following option is now per-listener:
* allow_anonymous
* password_file
* psk_file
* acl_file
* clientid_prefixes
* allow_zero_length_clientid
* auto_id_prefix
* auth_plugin
* auth_opt_*

To keep compatibility with existing config file, if any of those option is present before a listener directive, it will be used as global/default value. Having both global and per-listener works, but the per-listener override it.

Example:
```
acl_file acl-global
password_file pw-global
auth_plugin plugin-global.so

listener 1883
acl_file acl-listenner1
auth_plugin plugin-1.so
auth_plugin plugin-2.so

listener 1884
password_file pw-1884

listener 1885
```
Will result in three listeners:
* On port 1883:
  * acl_file from acl-listenner1
  * password_file from pw-global
  * auth_plugin plugin-1.so, then plugin-2.so
* On port 1884:
  * acl_file from acl-global
  * password_file from pw-1884
  * auth_plugin plugin-global.so
* On port 1885:
  * acl_file from acl-global
  * password_file from pw-global
  * auth_plugin plugin-global.so (the SAME instance as the one used by listener of port 1884)

Note:
* there is a breaking change: if someone specify auth options after a listener directive, it's no longer global.
* Override work per option. If a listener define password_file, only password_file is overridden. Another behavior could be that if any authentication/authorization option is defined in a listener, no global option are used for this listener (use case: use ACL file/password file by default, and for one listener use auth_plugin but not files).
* In the above example, auth_plugin used in global option is instantiated only once (e.g. call to plugin's mosquitto_auth_plugin_init & co) even if used by multiple listeners. This avoid a breaking change with plugin that does not support being instantiated multiple time (for example due to usage of some global variables).
* On the other hand if the same plugin is used within multiple listener directive, it will be instantiated multiple times.

Tasks to do:
* validate that the behavior for global/override is the one user expect.
* write some tests
* write some docs
* There is one gotcha with persistent session. Subscription of disconnected client/disconnected client don't store the associated listener, therefor when a disconnected client receive a message, an ACL check will be performed with no associated listener. In this case we can not perform the ACL check using the correct listener configuration and with this PR the global one is used. Note: currently mosquitto will perform another ACL check when the client reconnect, at this time the listener will be defined and correct ACL will be performed, assuming the global ACL check didn't rejected the message.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -66,6 +66,7 @@ static int conf__parse_int(char **token, const char *name, int *value, char *sav
 static int conf__parse_string(char **token, const char *name, char **value, char *saveptr);
 static int config__read_file(struct mosquitto__config *config, bool reload, const char *file, struct config_recurse *config_tmp, int level, int *lineno);
 static int config__check(struct mosquitto__config *config);
+static void conf__free_auth_plugins(struct mosquitto__listener *listener);
 
 static int conf__attempt_resolve(const char *host, const char *text, int log, const char *msg)
 {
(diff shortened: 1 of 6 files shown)


ralight:
This is now implemented in a different way. I wanted to preserve the option backwards compatibility with the config file.

I'm going to close this, thanks for your work on it though.

PR #567, opened 2017-09-28
Add a callback for manipulating the SSL context before connecting

joelw:
This patch adds a new on_ssl_ctx callback which allows client applications to manipulate the OpenSSL context prior to making connections. libcurl provides a similar facility with its ssl_ctx_callback function.

This can be used to do advanced SSL validation, or for my particular use case, to override the check date for certificate date validation when the system time() command cannot be used in an embedded environment.

Signed-off-by: Joel Williams <joel@joelw.id.au>

### DIFF
--- a/lib/cpp/mosquittopp.cpp
+++ b/lib/cpp/mosquittopp.cpp
@@ -63,6 +63,12 @@ static void on_log_wrapper(struct mosquitto *mosq, void *userdata, int level, co
 	m->on_log(level, str);
 }
 
+static void on_ssl_ctx_wrapper(struct mosquitto *mosq, void *userdata, void *ssl_ctx)
+{
+	class mosquittopp *m = (class mosquittopp *)userdata;
+	m->on_ssl_ctx(ssl_ctx);
+}
+
 int lib_version(int *major, int *minor, int *revision)
 {
 	if(major) *major = LIBMOSQUITTO_MAJOR;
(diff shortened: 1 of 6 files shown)


eslopez92:
Does this patch only apply when using the c/c++ libraries? Could this SSL CTX call back function be set using the LUA API? I have the following setup:

```
    ** LUA: Mosquitto Client **
require("ncr")     require("mosquitto")
    |                          |
    |                          |
 C to LUA                   C to LUA
 Bindings                   Bindings
    |                          |
   \/                         \/
C: NCR API                 Mosquitto
(create custom              Package
SSL CTX callback
function here)
```

I have an custom ncr package implemented in c with lua bindings which already leverages the libcurl ssl_ctx_function(). The problem is that it is in a separate LUA so module from the Mosquitto module, and both are unaware of each other. How could I tie that SSL CTX function to the mosquito from the C API to the LUA Mosquitto client to the C Mosquitto package?

karlp:
Emilio, you're going to have to try this out yourself.  Modify the lua bindings locally, with this patch applied, or port your test code to C and use this directly, and see if it does/does not meet your needs.

joelw:
Hi Emilio,

Sorry, I wasn't aware of any Lua bindings - if this is a separate project you would need to implement the new callback in those bindings too.

Here's an example of how to use the new callback:

```c++
void Mqtt::on_ssl_ctx(void *sslctx) {
    X509_VERIFY_PARAM_set_time( ((SSL_CTX * )sslctx)->param, get_time());
}
```
By default OpenSSL uses the time(NULL) function when checking certificate validity. This lets you set an SSL_CTX parameter to override this with your own time, for example if the system time is wrong and you do not have any control over it, but you do know the correct time from a call to an RTC device.

ralight:
Thank for this contribution (and sorry for the delay looking at it). I've just pushed a change that allows you to provide your own SSL_CTX rather than use the callback as you have done. Could you take a look and see whether it works for your needs? The changeset is 8470ca89b9ea4123e386455bb30bcd6f548792a4.

joelw:
Thanks, and no worries at all! Your approach sounds fine too.. I'll test my application against the new version when I have a chance and will let you know if there are any problems.

Issue #393, opened 2017-03-07
Mosquitto inflight_messages = 1, but got more than 1 message simultaneously


Asoul:
I've set `max_inflight_messages` to 1, but after stress test, my client got multiple simultaneous messages.

Does I misunderstood something? I thought `in-order delivery of messages` means every message should sent through PUBLISH, PUBREC, PUBREL, PUBCOMP, then next message.

Extracted `mosquitto.log`
```
1488858217: Sending PUBLISH to client (d0, q2, r0, m669, 'test/topic', ... (7 bytes))
1488858217: Sending PUBLISH to client (d0, q1, r0, m670, 'test/topic', ... (7 bytes))
1488858217: Sending PUBLISH to client (d0, q2, r0, m671, 'test/topic', ... (7 bytes))
1488858217: Sending PUBLISH to client (d0, q2, r0, m672, 'test/topic', ... (7 bytes))
1488858217: Sending PUBLISH to client (d0, q2, r0, m673, 'test/topic', ... (7 bytes))
1488858217: Received PUBREC from client (Mid: 669)
1488858217: Sending PUBREL to client (Mid: 669)
1488858217: Received PUBACK from client (Mid: 670)
1488858217: Received PUBREC from client (Mid: 671)
1488858217: Sending PUBREL to client (Mid: 671)
1488858217: Received PUBREC from client (Mid: 672)
1488858217: Sending PUBREL to client (Mid: 672)
1488858217: Received PUBREC from client (Mid: 673)
1488858217: Sending PUBREL to client (Mid: 673)
```

Thanks,
Asoul

ThomasKrannich:
Same behavior and question here. (Broker is Mosquitto and Subscriber is paho (Java))
QoS is 2 and the above behavior leads to messages getting lost - as it seems because retry_interval is reached before the first published message was confirmed with PUBREC.

asicoe:
I am seeing the same behaviour as Asoul.
My test is not even a stress test
I have a mosquitto broker and a publisher and a subscriber client. Both using java paho 1.1.1.
I publish some messages to one topic, then let the consumer consume for a while. Then shutdown the consumer (which has a persistent session). During the downtime some more messages are published. When the consumer is brought back mosquitto publishes multiple messages together exactly in the pattern that Asoul pasted. So we end up with multiple msgs in flight which leads to out of order messages in our app. This happens with Mosquitto 1.4.12 and max_inflight_messages = 1.

[...]

PR #568, opened 2017-10-01
Add test of QoS2 messages in-order with max_inflight_messages=1

toast-uz:
See #393


toast-uz:
Closing because original issue had been solved by releasing v1.5.

PR #572, opened 2017-10-03
Add websockets_headers_size option

### LINKED ISSUE #208: Websocket data will be truncated if HTTP header is too long
Since libwebsockets V1.7.0, it adds a **max_http_header_data** in info struct:

https://github.com/warmcat/libwebsockets/blob/master/changelog#L370

> - max_http_header_data: 0 for default (1024) or set the maximum amount of known
>   http header payload that lws can deal with.  Payload in unknown http
>   headers is dropped silently.  If for some reason you need to send huge
>   cookies or other HTTP-level headers, you can now increase this at context-
>   creation time.

Therefore if mosquitto's websocket interface receives a HTTP header over 1024 bytes, all additional bytes will be dropped. Maybe it is better to make this value configurable at mosquitto config file?

PierreF:
This PR fix #208 by adding an option websockets_headers_size which is passed to libwebsockets max_http_header_data.

### DIFF
--- a/man/mosquitto.conf.5.xml
+++ b/man/mosquitto.conf.5.xml
@@ -784,6 +784,18 @@
 							Defaults to 0.</para>
 					</listitem>
 				</varlistentry>
+				<varlistentry>
+					<term><option>websockets_headers_size</option> <replaceable>size</replaceable></term>
+					<listitem>
+                        <para>Change the websockets headers size. This is a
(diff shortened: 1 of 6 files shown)


ralight:
This has now been merged as 1aaf5f2348d437a1f18a8a7349fb0027ae6ab242, thanks for your patience.

PR #575, opened 2017-10-04
Ipv4 fallback 574

kevinhgardner:
If a connection to an IPv6 network and it fails, then on the next attempt try to connect to an IPv4 address instead.


kevinhgardner:
Should have been for 'fixes'-branch instead of 'master'-branch.

Issue #574, opened 2017-10-04
Should fall back to IPv4 if IPv6 fails.


kevinhgardner:
If the mqtt host resolves to both IPv6 and IPv4 addresses, and there is a routing issue over IPv6, mqtt should fall back to IPv4.
See http://dev.eclipse.org/mhonarc/lists/mosquitto-dev/msg01633.html for more details.

PR #576, opened 2017-10-04
If a connection to an IPv6 network and it fails, then on the next att…

kevinhgardner:
…empt try to connect to an IPv4 address instead.

Bug: https://github.com/eclipse/mosquitto/issues/574

Signed-off-by: Kevin Gardner <kevinhgardner@gmail.com>

### DIFF
--- a/lib/mosquitto_internal.h
+++ b/lib/mosquitto_internal.h
@@ -272,6 +272,10 @@ struct mosquitto {
 	UT_hash_handle hh_sock;
 	struct mosquitto *for_free_next;
 #endif
+	/* If set, ignore ipv6 addresses.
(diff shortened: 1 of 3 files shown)


toast-uz:
Hi,
This seems a new feature and not a fixing a bug, so you should create a new branch from the develop.

kevinhgardner:
Hi,
I would more consider this a bug since it breaks the current functionality.
That's why i submitted it on the fixes branch.
If you prefer I submit it on develop, I can change it.

toast-uz:
I like it on develop. Many PRs on develop broke the current functionality.
Thanks.

kevinhgardner:
Moved to develop:
https://github.com/eclipse/mosquitto/pull/578

PR #578, opened 2017-10-05
If a connection to an IPv6 network and it fails, then on the next att…

kevinhgardner:
If a connection to an IPv6 network and it fails, then on the next attempt try to connect to an IPv4 address instead.
Bug: #574

### DIFF
--- a/lib/mosquitto_internal.h
+++ b/lib/mosquitto_internal.h
@@ -278,6 +278,10 @@ struct mosquitto {
 #ifdef WITH_EPOLL
 	uint32_t events;
 #endif
+	/* If set, ignore ipv6 addresses.
+	   This variable will be set to true when a connection to an ipv6 address is attempted.
+	   If the connection is successfully set-up, the value will be reset. */
+	bool ignore_ipv6;
 };
 
 #define STREMPTY(str) (str[0] == '\0')
--- a/lib/net_mosq.c
+++ b/lib/net_mosq.c
@@ -326,6 +326,9 @@ int net__try_connect(struct mosquitto *mosq, const char *host, uint16_t port, mo
 	}
 
 	for(rp = ainfo; rp != NULL; rp = rp->ai_next){
+		/* Only ignore ipv6 if there is another entry in the list */
+		if (mosq->ignore_ipv6 && rp->ai_family == PF_INET6 && NULL != rp->ai_next) continue;
+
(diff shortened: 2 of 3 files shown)


dirkfeytons:
Any feedback on this patch?
Are the failing tests blocking the merge or is other work needed? Or is the functionality as such not deemed useful?

lib/mosquitto_internal.h
```diff
@@ -278,6 +278,10 @@ struct mosquitto {
 #ifdef WITH_EPOLL
 	uint32_t events;
 #endif
+	/* If set, ignore ipv6 addresses.
+	   This variable will be set to true when a connection to an ipv6 address is attempted.
+	   If the connection is successfully set-up, the value will be reset. */
+	bool ignore_ipv6;
```
karlp: seems like a bad variable name then.  try_ipv4_first_next?  skip_v6_next?

lib/net_mosq.c
```diff
@@ -326,6 +326,9 @@ int net__try_connect(struct mosquitto *mosq, const char *host, uint16_t port, mo
 	}
 
 	for(rp = ainfo; rp != NULL; rp = rp->ai_next){
+		/* Only ignore ipv6 if there is another entry in the list */
+		if (mosq->ignore_ipv6 && rp->ai_family == PF_INET6 && NULL != rp->ai_next) continue;
```
karlp: So now you're goign to try the _last_ ipv6 address if there's only ipv6 addresses, instead of the first one?  I don't really think that's what the resolver intends when it gives you a list.

karlp:
Personally this smells bad.  Can't you fix this with your own dns?  Trying to hack aroudn it like sounds very special case, surprising and difficult to diagnose what it's actually doing.  It's not proper round-robin, it's something oddly inbetween, except not really.

Do you realllllly just want that, "when a name returns multiple addresses, if one fails, try the next one next time" ?  That sounds more understandable, easier to document, and more likely to be what people actually want.

rlfnb:
https://docs.oracle.com/javase/7/docs/technotes/guides/net/ipv6_guide/index.html#ipv6-related from my POV, if ipv6 resolves to an address, the address has to be functional. The ipv6 properties are made for such use cases and work for each use case in the java world. Introducing some non-java magic won't make it easier to cope with the implementation.

ralight:
This was closed by accident.

@rlfnb I think you're arguing against this change, but I'm not 100% sure. Can you elaborate please?

Issue #539, opened 2017-09-01
Bridge Enhancement: Add a writable $SYS/broker/bridge/disconnect topic for [temporarily] disconnecting a bridge connection.


cjmaynard:
I posted a query about this and didn't receive any negative feedback about the idea, so I’m opening an enhancement request here in the hope that it could be implemented.  My post is here: http://dev.eclipse.org/mhonarc/lists/mosquitto-dev/msg01611.html.

Basically, I have a bridge that can connect to more than one possible remote broker, each one reachable via a separate interface.  I would like to be able to force the bridge to close its current connection under certain conditions so that it can establish a new connection with the alternate remote broker much quicker than it otherwise would.

The specific use case that I have is that if I’m able to detect that an interface is down (such as if the Ethernet cable becomes disconnected), I would like to cause the bridge to immediately abandon its current connection if it happens to be communicating with a remote broker over that particular interface so that it can more quickly establish a connection with the other remote broker over the other interface, rather than having to wait for timers to expire, which takes much longer to occur.  It is possible to reduce timeout values somewhat, such as the keepalive value, but it’s still not responsive enough and has the unfortunate side-effect of introducing more frequent traffic, which we’d also like to avoid.  It’s important for us to remain connected as much as possible in the event data needs to be sent, but we need to keep data usage as low as possible when we’re not actively transmitting data.  I would think data usage would be a common concern for many people?

I’ve seen some recent activity surrounding issue #535 and while I don’t think the resolution of that issue will directly solve this one, there does seem to be interest in being able to control the broker via writable topics, so in keeping with issue #535, I might propose a writable topic of `$SYS/broker/bridge/disconnect` as a possible solution to this issue.  I’m open to other possible alternatives that accomplish the same goal though.  I’m also fine with adding “`allow_sys_update true`” to the`mosquitto.conf` file, as described in issue #535.

toast-uz:
Hi,
Different from what you want to do, FYI, I tested optimization of current bridge timers in the dual bridge uplink configuration.

Default setting:
- Detecting the disconnection of primary bridge uplink: 60s
- Connecting alternative: 31s
- total: 91s
- background traffic: 0.5 bps/connection (2byte PING * bi-direction / 60s)

Optimized setting:
- Detecting the disconnection of primary bridge uplink: 5s (by setting keepalive_interval 5)
- Connecting alternative: 2s (by setting restart_timeout 1)
- total: 7s
[...]

PR #584, opened 2017-10-12
 Added new signal SIGWINCH to signal broker, that it should try to connect to disconnected bridge(s) now.

jsaak:
```
The SIGWINCH signal causes mosquitto to try to
reconnect to all bridges, which are disconnected at the
moment. It is useful when the internet connection is weak.
As soon as you have internet or vpn connectivity, you can
signal the broker, to try to reconnect. Set start_type to
auto (the default), and you can give a big number to
restart_timeout. This way as soon as you have internet,
you can reconnect with the bridge

Since there is no SIGUSR3, I picked one which is rarely used.
```

### DIFF
--- a/man/mosquitto.8.xml
+++ b/man/mosquitto.8.xml
@@ -452,6 +452,18 @@
 					testing feature only and may be removed at any time.</para>
 				</listitem>
 			</varlistentry>
+			<varlistentry>
+				<term>SIGWINCH</term>
+				<listitem>
+          <para>The SIGWINCH signal causes mosquitto to try to
+          reconnect to all bridges, which are disconnected at the
+          moment. It is useful when the internet connection is weak.
+          As soon as you have internet or vpn connectivity, you can
+          signal the broker, to try to reconnect. Set start_type to
+          auto (the default), and you can give a big number to
+          restart_timeout.</para>
+				</listitem>
+			</varlistentry>
 		</variablelist>
 	</refsect1>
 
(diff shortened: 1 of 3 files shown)


toast-uz:
Hi,
I am not a committer but I'd like to ask something.
- Could you show some examples of major softwares use SIGWINCH for such an unintended use?
- Could you show your opinion comparing #535? I guess implementation with extension of the method of #535 may be better than using unintended SIGWINCH.

jsaak:
SIGWINCH is a hack. I do agree that it is not a good solution.
Publishing to $SYS/broker/bridge/reconnect (or any other $SYS topic) would be perfectly fine.

toast-uz:
Thanks.
FYI, look at #539 as a similar requirement.

toast-uz:
@jsaak, could you close this PR as a duplicate of #535 ?

jsaak:
I am not sure. Does that PR include the functionality to trigger a reconnect?
I need to trigger a reconnect. #539 wants to trigger a disconnect. Is publishing to $SYS topic now a merged feature? If yes then both this and #539 should be trivial to implement.

toast-uz:
I think your request is on the road beyond #535 , and I believe you agreed your implementation was not fully appropriate and  #535 was a better one.
Therefore, would you add a comment on #535 rather than keeping this independent PR ?

fcarlier:
on #653 ;)

Envoyé de mon terminal mobile

toast-uz:
Sorry #535 was closed and #653 is available now. Thanks @fcarlier !

jsaak:
Ok it seems that you are aware of the situation. I will close this. You have the link with the implementation of the feature, if you need it. (this PR). Just pls do not forget to add the reconnect topic :)

Issue #275, opened 2016-09-26
mosquitto_sub timeout.


HariniJSrowa:
Does mosquitto_sub have the feature of waiting for a response and if it does not get any response it should timeout within a time frame.

ralight:
No it doesn't. You can specify that it exits after a certain count of messages though. I'm going to leave this open as a possible enhancement.

HariniJSrowa:
Thanks Roger ! Well, the test case is with a setup of all three mosquitto_pub, mosquitto_sub and mosquitto broker running on one single machine with either CentOS 6.5/Ubuntu 14.04 LTS. I run mosquitto_pub in a while loop (forever)and  keep publishing data on a certain topic and subscribe too (forever)for the same. I terminate mosquitto_pub after some time and that leaves mosquitto_sub hung :(. I want mosquitto_sub to quit or receive some event from broker that there are no more publishers running but mosquitto_sub just keeps sending PINGREQ and receives PINGRESP as per the protocol.

Do you suggest any way for the mosquitto_sub to know that there are no more messages to be received since the publisher itself is not running? In this case it cannot even exit after the count of messages since if I run mosquitto_sub -h 127.0.0.1 -t Test -C 1 it hangs since no publisher is running.

ralight:
At the moment, no this is not possible. This would be quite trivial for you to implement by writing your own client though.

HariniJSrowa:
Yes ! Thanks !

victorclaessen:
So sorry to resurrect this thread (but it was never closed, it seems). This is a really useful feature. For instance, for checking to see if an 'alive' signal is received within a certain time frame.

I was just wondering, would it be useful to set the exit code of the mosquitto_sub process according to whether there was a message received or not? Perhaps 0 if there was a message, and 1 if the timeout had occurred? (I realise that one could also check stdout to see if it was empty or not.)

ralight:
The timeout option was added a while back in 1.6. I've just added the option to have a timeout specific exit code, which will be in 1.7.

victorclaessen:
Awesome. Thanks!

PR #589, opened 2017-10-17
mosquitto_sub: add simple timeout for message processing (refs #275)

mhei:
This PR implements a simple timeout handling while processing MQTT messages
in sub client. The implementation is based on SIGALRM, so to avoid conflicts with
sleep(3) calls (which could be implemented also using SIGALRM), the sleep(3)
usage in mosquitto_loop_forever is factored out and replace with nanosleep(2)
in a first step.


ralight:
I do like this, thanks for taking the time to submit it and my apologies for taking so long to look at it.

PR #590, opened 2017-10-17
Quote ${DESTDIR} usage in various Makefiles

mhei:
This prevents failures during make install when whitespace
is part of DESTDIR pathname.


ralight:
Thanks, this looks fine to me. Sorry for the delay merging it.

PR #591, opened 2017-10-17
readme: mention documentation build requirements

mhei:


PierreF:
I think the reason why xsltproc is not in build dependencies is that it's only required when building from git. When building from released source, the man page is already generated and xsltproc is not required.
If adding it to build dependencies paragraph, I think it may specify that this requirement is only for build from git source (note that in it what is say in the last sentence of Building from source paragraph).

ralight:
@PierreF is right, but I think it's still helpful to have this information there so thanks for your contribution! I've added a short comment about it only being needed when building from git.

Issue #457, opened 2017-06-08
docker: Error response from daemon: readdirent: not a directory.


robmarkcole:
Attempting to run the docker image https://hub.docker.com/_/eclipse-mosquitto/ on docker on my Synology using `sudo docker run -it -p 1883:1883 -p 9001:9001 -v mosquitto.conf:/mosquitto/config/mosquitto.conf -v /mosquitto/data -v /mosquitto/log eclipse-mosquitto` I get the error: `docker: Error response from daemon: readdirent: not a directory.`

ralight:
I'm relying on others for docker support - any suggestions?

Sammers21:
I have the same problem

dimarzio:
Me and one another colleague have the same problem!

cyclingengineer:
Seeing the same thing on debian jessie.

kainoj:
Something is messed up with `-v` flag.  If you don't need custom configs simply remove the flag:

```bash
sudo docker run -it -p 11883:1883 -p 9001:9001 eclipse-mosquitto
```
If you need, do the following:
- create `/mosquitto/config/`, `/mosquitto/data/`, `/mosquitto/log/` dirs on your host machine
- put `mosquitto.config` in `/mosquitto/config/`. Don't forget about those options in config:

```
persistence true
persistence_location /mosquitto/data/
log_dest file /mosquitto/log/mosquitto.log
```

Now just simply execute
```bash
# Notice that mosquitto is exposed on 11883 on host
sudo docker run -it -p 11883:1883 -p 9001:9001 -v /mosquitto/:/mosquitto/ eclipse-mosquitto
```

It all worked on a fresh Ubuntu Server 16.10 vm.

robmarkcole:
Getting a very cryptic error on `sudo docker run -it -p 1883:1883 -p 9001:9001 eclipse-mosquitto`

`docker: Error response from daemon: driver failed programming external connectivity on endpoint desperate_austin (de61691ef....): Error starting userland proxy: listen tcp 0.0.0.0:1883: bind: address already in use.`

tompscanlan:
I hit the same problem. 

I copied the example from https://hub.docker.com/_/eclipse-mosquitto/, so 
 I was running `-v mosquitto.conf:/mosquitto/config/mosquitto.conf:ro` which failed.  Changed it to `-v $(pwd)/mosquitto.conf:/mosquitto/config/mosquitto.conf:ro` and it's working. 

[...]

PR #595, opened 2017-10-23
add full path to config file to fix docker run error

tompscanlan:
This should prevent cut-n-paste of the readme from  causing issue #457.  Should also get patched in the coverity-develop branch docker readme.


tompscanlan:
duh... will follow the contributor doc :)

Issue #608, opened 2017-11-05
TLS Error during connect, hard to reproduce


JonoJensen:
When adding TLS Ticket support a hard to reproduce issue appeared for me. The issue can be seen in server loggs as: 
`1509547052: OpenSSL Error: error:1408F10B:SSL routines:SSL3_GET_RECORD:wrong version number
`

It can be observed with strace as the pressence of a line similar to (In my case the length - 291 - could be used as a catch syscal condition to GDB):
`write(5, "\26\3\1\1\36\1\0\1\32\3\3\272\221\10?+\37\t\2\270^c\211\24p\376\202A\311#\312\212"..., 291) = 291
`
 
This appears in a tls1.2 connection after encryption has been turned on. The 291 bytes write is a new client_hello, sent with version tls 1.0 - thus the error message on the server.

Reason is that mosquitto client tries to transmitt a frame before SSL_connect has finnished (has reported _WANT_READ). The data emmited apperas on the following call to SSL_connect. One way to reproduce is to let multiple mosquitto_sub connect with same client certificate to samer server on localhost kicking each other out. Depending on timing one of the clients may fail after a number of reconnections.

Fix suggestion (maybe ther's a better one, this works for me. )
I have not searched for other places where a SSL_x() call may be made prior to SSL_connect() has finnished if interrupted by a _WANT_READ.

```
From 10b9023261ce948b07a09ab5ab7515079b2f5c85 Mon Sep 17 00:00:00 2001
From: Jon Jensen <jon.jensen@permobil.com>
Date: Fri, 3 Nov 2017 22:08:02 +0100
Subject: [PATCH] Fix issue when SSL_connect() returns SSL_ERROR_WANT_READ. A
 call to SSL_write here will transmit a new client hello and make ssl
 connection fail.

---
 lib/net_mosq.c | 4 ++++
 1 file changed, 4 insertions(+)

diff --git a/lib/net_mosq.c b/lib/net_mosq.c
index a1f1e6f..814224a 100644
--- a/lib/net_mosq.c
+++ b/lib/net_mosq.c
@@ -935,7 +935,11 @@ int _mosquitto_packet_write(struct mosquitto *mosq)
 	}
 	pthread_mutex_unlock(&mosq->out_packet_mutex);
 
+#if defined(WITH_TLS) && !defined(WITH_BROKER)
+	if((mosq->state == mosq_cs_connect_pending)||mosq->want_connect){
+#else
[...]

PR #609, opened 2017-11-06
Fix issue when SSL_connect() returns SSL_ERROR_WANT_READ. A call to S…

JonoJensen:
Fix issue when SSL_connect() returns SSL_ERROR_WANT_READ. A call to SSL_write here will later transmit a new client hello and make ssl connection fail.

Se issue #608


JonoJensen:
Incorrect author, I remove this and make a new attempt.

PR #611, opened 2017-11-06
Fix issue when SSL_connect() returns SSL_ERROR_WANT_READ.

JonoJensen:
Fix issue when SSL_connect() returns SSL_ERROR_WANT_READ. A call to SSL_write here will later transmit a new client hello and make ssl connection fail.

Se issue #608#

Signed-off-by: JonoJensen <jono.jensen@yahoo.se>


ralight:
This is a good catch. Thanks very much for your contribution!

PR #612, opened 2017-11-06
Dev client ticket support

JonoJensen:
Adds TLS ticket support to client lib and test clients.
Adds control over TLS ticket life length to server.
(Does not reissue tickets on same session to force renegotiation when time expires.)
Requires issue #608 fix.


JonoJensen:
There are 2 tests failing for make test. These however failed on the branch prior to this patch at least on my system./Jon

Issue #537, opened 2017-08-30
Setting "connection_messages false" still logs a "New connection from X on port Y." message


cjmaynard:
Regarding the connection_messages option, the [mosquitto.conf man page](https://mosquitto.org/man/mosquitto-conf-5.html) reads:

       connection_messages [ true | false ]
           If set to true, the log will include entries when clients connect
           and disconnect. If set to false, these entries will not appear.

With the default of "connection_messages true", and using mosquitto_pub to test, I see messages like so:

    1504137323: New connection from 127.0.0.1 on port 1883.
    1504137323: New client connected from 127.0.0.1 as mosqpub|2501-test (c1, k60).
    1504137323: Client mosqpub|2501-test disconnected.

... but with "connection_messages false", I still see this message:

    1504137889: New connection from 127.0.0.1 on port 1883.

I would not expect to see that message when setting "connection_messages false".

karlp:
the variable name is perhaps bad, in that it says connections, but the manual text says it's the logs on clients, and that's true?  I'd probably generally agree though, that it's not the ideal behaviour

toast-uz:
+1
I agree this option is not implemented  consistently.

PR #613, opened 2017-11-06
Fix for issue #537: should not log connection message if option disabled

szarta:
Added configuration check before printing connection message.

Signed-off-by: Brandon Arrendondo <barrendo@gmail.com>


szarta:
I will test and find the other ones as well and include those in the pull request.

szarta:
The help for connection_messages logging states that it should affect all messages relating to clients connecting and disconnecting, so I found those and made sure they were wrapped appropriately.

PierreF:
I'm not convinced that all messages are *connection* related messages. The ACL denied message don't look connection related for me.
The message about expiring persistent also is not connection related in my opinion.

PR #614, opened 2017-11-08
Required to use r2 for the dockerfile to work

agrosjea:
Hello, here is a small update as r0 was not working anymore.


agrosjea:
Mmmm sorry that pull request isnt valid at all...

Let me correct that

PR #617, opened 2017-11-08
docker 1.4.14 using r2

agrosjea:
Signed-off-by: agrosjea <arnaudg@momentfactory.com>


ralight:
Thanks for your work on this, I'm closing it because the docker file has been updated.

Issue #573, opened 2017-10-03
Brokers Dies on Windows (512 connections)


ericGTT:
When I have 510-512 connections the broker dies. 
Running on Windows (The documentation I read said 1024 unless running on Windows Server.)
I tried both Windows Server and Windows 7.
And using TLS.
Mosquitto Version 1.4.14
Running as a service

-- Update --
I found out if I set " _setmaxstdio(2048);" and compile code I can get more connections. Most I've had was about 2200+ connections. (I'm still testing it.)

ralight:
Closing as per merged PR.

PR #624, opened 2017-11-14
Increase maximum connections (Windows) from 512 to 2048 (Maximum)

ericGTT:
#573 
Increase maximum connections using _setmaxstdio for windows.


ralight:
I've just pushed a commit including this to the develop branch, but doing it ever so slightly differently, so I'm going to close this. Thanks for your contribution!

ralight:
I'm putting it on fixes as well for the 1.4.16 release.

PR #625, opened 2017-11-17
Enable notice for PINGRESP

redmember:


karlp:
Could you provide some sort of rationale for this?  It's information that is already available at debug level, and is going to emit log lines for every client, for normal operation of the protocol.  It's an incredible volume of messages for a lot of clients.  Leaving it at debug seems entirely appropriate.

PR #635, opened 2017-12-02
Wrong Docbook Path

TheBeachMaster:
Fixed an issue where Mosquitto would search for the wrong Docbook path during build.


toast-uz:
Could you explain what exactly the issue is?
According to readme.md, section "Building from source", we cannot compile docs except for on Debian/Ubuntu systems.

ralight:
As @toast-uz suggests, the path is correct for Ubuntu/Debian based systems which is what I develop on. I'd be happy to take a solution that meant it works for more systems.

TheBeachMaster:
Noted... 
This issue came about when building on Centos 7. 
Will it be okay if I updated the docs instead? and probably point out that the path to *docbooks* is different for Centos systems.

toast-uz:
It will be a good solution If you can update man/Makefile to switch building on each OS.

TheBeachMaster:
Great idea @toast-uz . Let me look into it

PR #651, opened 2017-12-10
fix reproducible builds by removing build timestamp

lynxis:
Build timestamps prevents reproducible builds. [0]

[0] https://reproducible-builds.org/docs/timestamps/

Signed-off-by: Alexander Couzens <lynxis@fe80.eu>

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -309,7 +309,7 @@ void mqtt3_config_cleanup(struct mqtt3_config *config)
 
 static void print_usage(void)
 {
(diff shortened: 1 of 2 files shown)


Tifaifai:
Integrate this PR in "Bridge dynamic by a publish on $SYS topic." #653

ralight:
Thanks for the suggestion, I hope you don't mind I've made a more complete change and added it to the develop branch instead.

karlp:
updated version looks even better! :+1:

PR #692, opened 2018-02-07
Create 'setup.sh' to help setting up environment

tjhuang:
Replace the 'travis-configure.sh' and 'travis-install.sh' to 'setup.sh' for system tools installation on both Linux system and MacOS.

Signed-off-by: Jacob Huang <tjhuang.lec@gmail.com>


ralight:
Sorry, but this completely breaks travis integration which is the whole point of those scripts existing.

PR #698, opened 2018-02-14
fix incorrect return code when connecting in non-blocking mode

krismattheus:
Problem: When a broker is configured to make a bridge connection, the bridge connection is no longer established.
Platform: Windows

When the broker is configured to make a bridge connection, the _mqtt3_bridge_connect_ function fails, because the non-blocking __mosquitto_send_connect_ call returns an error (WSAENOTCONN=10057).
The error happens because the socket connection is still pending, but the __mosquitto_socket_connect_ call doesn't return MOSQ_ERR_CONN_PENDING in this case.
Apparently, the return code is overwritten in __mosquitto_socket_connect_.

### DIFF
--- a/lib/net_mosq.c
+++ b/lib/net_mosq.c
@@ -645,7 +645,7 @@ int _mosquitto_socket_connect_step3(struct mosquitto *mosq, const char *host, ui
 int _mosquitto_socket_connect(struct mosquitto *mosq, const char *host, uint16_t port, const char *bind_address, bool blocking)
 {
 	mosq_sock_t sock = INVALID_SOCKET;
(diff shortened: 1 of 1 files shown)


krismattheus:
First I had only tested this with an insecure connection (port 1883), because only this setup was giving connection errors.
I've retested it with a TLS connection to a public broker, and both the existing and modified code are connecting OK.
Note that this was tested with version 1.4.15.

ralight:
This has now been merged as 479d8e5f1a162927d5e67e2bdd754bba880f28a9, thanks for your patience.

PR #712, opened 2018-03-04
Cmake fixes

aconstlink:
- Setting default values on WIN32 for CMake options so Mosquitto compiles out of the box
- Changed c-ares include behaviour making CMake using find_package from c-ares CMake build env
- Added arpa/nameser.h dummy for successful Visual Studio Win32 build


karlp:
If you want anyone to review this, you're probably going to have to rebase this on top of master, rather than include all of the other changes here that are out of scope.

aconstlink:
Ok. Sorry. Is it required to somehow delete the pull request?

PR #713, opened 2018-03-04
CMake fixes for Windows build

aconstlink:
- by default disabling WITH_SRV, WITH_THREADING and SSL support for out of the box building on WIN32 Visual Studio
- fixed c-ares CMake search with find_package function which works with current c-ares CMake build and its Cmake configs
- added arpa/nameser.h dummy for WIN32 for successful compilation with c-ares search ares_search in srv_mosq.c


karlp:
This is now much more legible and plausible to review :) don't worry about the old one, just leave it closed.

aconstlink:
I signed that ECA thing, but it still shows as failed

ralight:
I support the idea of changing some defaults - DNS SRV is set to off by default for all platforms in 1.5 anyway. I would argue that TLS support should be in all builds, so you should go off and find openssl, but I'm willing to be persuaded on that point.

What is the origin of the nameser header? You should check to see whether that part of the contribution meets the ECA that you have just signed: https://www.eclipse.org/legal/ECA.php

Another point - you have used "( BLAH )" instead of "(BLAH)" throughout the parts you have contributed and your more recent commit goes on to modify all of the existing code to this style. Please do not do this, copy the existing style.

You also need to remember to add a "signed-off-by" line to your commit.

aconstlink:
Hi there.

That's great. I think, disabling all the mentioned options by default, gives a nice "out of the box" build, which is create for Visual Studio users on Windows at the moment.  Everyone who wants to have some more security/threading is totally able to opt-in all those options.

Sorry for the "low-brainer" spaces CMake changes. I thought CMake is kinda abandoned, since there is so much potential left, especially on Windows with the install target not working on Windows or no CMake config scripts, which would totally make Mosquitto super duper easy to build on Windows at least. I also changed it, because some functions where written like func (ABC) and others like func(ABC).

The header file is a little bit unfortunate, because there are only two enums needed from that header in order to work with c-ares on Windows.

I am totally eager to make more changes. Do you really want me to drop the last commit?

aconstlink:
Ok, done. I just dropped the last commit. Ill go from the original then.

aconstlink:
Also killed the first commit and recommited with "signed-off-by" commit option. By removal of the first commit, the request closed...

aconstlink:
I realized, that 1.5 is in the develop branch. Then I would rather proceed working there. So I close this one.

PR #725, opened 2018-03-13
websockets.c: unbreak build without TLS

jacmet:
Commit 7943072b1f3b (Fix use_identity_as_username not working on websockets
clients) added code which unconditionally accesses mosq-ssl, breaking the
build when TLS support is disabled.

Fix it by guarding this logic inside #ifdef WITH_TLS.

Signed-off-by: Peter Korsgaard <peter@korsgaard.com>


ralight:
That's great, thanks for your contribution!

PR #729, opened 2018-03-19
Use config from user writable snap directory

8none1:
This PR allows the loading of config from a user writable area on disk.  $SNAP_USER_COMMON is not versioned on upgrades, and so the user's config will remain the same across upgrades.

I've created a new "default_config.conf" file based on the config that was being shipped inside the snap directory (and removed the mosquitto.conf file).  The launcher script will fall back to this file if there isn't a user config file in the specified location.  This should mean that when run, the snap will actually do something useful.  I dropped "user root" from this example, as I don't think that it needs to run as root, but I might be wrong.  It seems to work ok here.

The launcher script first copies the example config (from the root of the source) in to the $SNAP_USER_COMMON directory, so that people can find it and read it, copy it, etc.  The launcher only does this if the file doesn't already exist. The launcher also tells the user where it is, so they'll hopefully be able to find it again later.

The launcher then looks in the $SNAP_USER_COMMON dir for "mosquitto.conf".  If it finds one it will pass it to mosquitto to use.  If not, it will fall back to using the default one included in the snap dir in source (and in the application's $SNAP dir on the machine it is installed on).  It will also echo some hints on what's going on so the user can understand where the config is being loaded from.

Launcher scripts like this are fairly standard practice where you need to access files at run time that can't exist at build time - because we can't be sure where on disk the will end up when installed.

It might be worth setting the CWD to $SNAP_USER_COMMON so that the persistence database gets written there as well, but I haven't made that change.  Should be simple enough if required.

### DIFF
--- a/snap/default_config.conf
+++ b/snap/default_config.conf
@@ -1,0 +1,2 @@
+port 1883
+persistence false
--- a/snap/launcher.sh
+++ b/snap/launcher.sh
@@ -1,0 +1,31 @@
+#!/bin/sh
+# Wrapper to check for custom config in $SNAP_USER_COMMON and use it
+# otherwise fall back to the included basic config which will at least
+# allow mosquitto to run and do something.
+# This script will also copy the full example config in to SNAP_USER_COMMON
+# so that people can refer to it.
+
+CONFIG_FILE="$SNAP/default_config.conf"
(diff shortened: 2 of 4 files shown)


8none1:
I've now signed the contrib agreement and added a signed-off-by.

ralight:
This is great, thanks very much for the detailed description.

PR #735, opened 2018-03-21
Makefile modifications to allow compilation agains specific sysroot

dezquerra:
Added variable SYS_ROOT to Makefiles to allow compilation against a specific sysroot. This variable is useful when crosscompiling using a specific or 3rd party toolchain with a defined or prefixed sysroot. The usage is the same as variable CROSS_COMPILE, already present

Signed-off-by: Dani Ezquerra <dezque@gmail.com>

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -9,25 +9,25 @@ static : static_pub static_sub
 	# libmosquitto only.
 
 static_pub : pub_client.o client_shared.o ../lib/libmosquitto.a
-	${CROSS_COMPILE}${CC} $^ -o mosquitto_pub ${CLIENT_LDFLAGS} -lssl -lcrypto -lpthread
+	${CROSS_COMPILE}${CC} --sysroot=${SYSROOT} $^ -o mosquitto_pub ${CLIENT_LDFLAGS} -lssl -lcrypto -lpthread
 
 static_sub : sub_client.o client_shared.o ../lib/libmosquitto.a
-	${CROSS_COMPILE}${CC} $^ -o mosquitto_sub ${CLIENT_LDFLAGS} -lssl -lcrypto -lpthread
+	${CROSS_COMPILE}${CC} --sysroot=${SYSROOT} $^ -o mosquitto_sub ${CLIENT_LDFLAGS} -lssl -lcrypto -lpthread
 
 mosquitto_pub : pub_client.o client_shared.o
-	${CROSS_COMPILE}${CC} $^ -o $@ ${CLIENT_LDFLAGS}
+	${CROSS_COMPILE}${CC} --sysroot=${SYSROOT} $^ -o $@ ${CLIENT_LDFLAGS}
 
 mosquitto_sub : sub_client.o client_shared.o
-	${CROSS_COMPILE}${CC} $^ -o $@ ${CLIENT_LDFLAGS}
+	${CROSS_COMPILE}${CC} --sysroot=${SYSROOT} $^ -o $@ ${CLIENT_LDFLAGS}
 
 pub_client.o : pub_client.c ../lib/libmosquitto.so.${SOVERSION}
-	${CROSS_COMPILE}${CC} -c $< -o $@ ${CLIENT_CFLAGS}
+	${CROSS_COMPILE}${CC} --sysroot=${SYSROOT} -c $< -o $@ ${CLIENT_CFLAGS}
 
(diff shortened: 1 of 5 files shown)


Tifaifai:
Where is the initialization of the variable ${SYSROOT} ? with CROSS_COMPILE ? .... in define by developper in config.mk file ?

dezquerra:
As well as CROSS_COMPILE variable, it is not initialized. If nothing is specified the default root filesystem will be used to compile mosquitto. Otherwise, if you give it a value on compilation time, e.g make CROSS_COMPILE=arm-poky-linux-gnueabi SYSROOT=/opt/poky/sysroot; it will search libraries and includes on the specified filesystem

Tifaifai:
Yes of course ...
Maybe propose with your commit, in config.mk : 
```
CROSS_COMPILE :=
SYSROOT :=
```

karlp:
or you know, fix your busted toolchain, so you don't have to go running around trying to  patch every application you ever try and build :)

karlp:
How exactly does this work when the variable is empty?  You're hard passing the argument with a blank option.

dezquerra:
As said in my first comment, if it's left empty it uses the default sysroot, it is the one of the host machine.

About the "busted toolchain". Sometimes, working with openembedded or Yocto the options to modify the toolchain are bit tricky. For some reference:

https://lists.yoctoproject.org/pipermail/yocto/2014-January/018023.html

As long as mosquitto is designed to work on IoT devices, we should assume that it could be needed for devices using old library versions, minimal library versions or one of the two projects mentioned above. So I think, that is not about fix a "busted toolchain" but ease compilation.

karlp:
So you're trying to have one person build your yocto toolchain, then you're trying to move it to another machine in a different location, and _then_ you're trying to patch external applications to support that you moved a toolchain that has static sysroots?  I still say you're making noisy pointless changes.

dezquerra:
1. Is quite common to have a person/enterprise building the toolchain for you, as well as the kernel
2. When you receive a toolchain it is usually a .sh, just like when you generate it
3. When you install a toolchain a default location is prompted. You can choose to use it or ignore it.
4. When you are not in control of your kernel, is quite common to specify the sysroot location, at least in my experience. May be because it is different or it is not specified at all or it is specified as /non/existent

ralight:
I don't believe this is necessary. You should just be able to do this:

`make CFLAGS=--sysroot=/path LDFLAGS=--sysroot=/path`

dezquerra:
Ralight you are right. The CFLAGS will do the same if used together with make, however as we can see on the Makefiles, CFLAGS variable is not considering when building most of the targets. 

Another option will be to add CFLAGS instead of SYSROOT that offer most options that the current added sysroot.

karlp:
As the readme says, cmake is for windows builds.  You should be using the makefiles as documented.

dezquerra:
karlp can you please clarify your comment? There is nothing about cmake in this discussion.

ralight:
@dezquerra sorry but you are mistaken, CFLAGS is passed to every .c compilation and LDFLAGS is passed to every linker step. Try it out to see.

dezquerra:
Hi @ralight, from the files modified, i see:

static_pub : pub_client.o client_shared.o ../lib/libmosquitto.a
-	${CROSS_COMPILE}${CC} $^ -o mosquitto_pub ${CLIENT_LDFLAGS} -lssl -lcrypto -lpthread
+	${CROSS_COMPILE}${CC} --sysroot=${SYSROOT} $^ -o mosquitto_pub ${CLIENT_LDFLAGS} -lssl -lcrypto -lpthread
 
 static_sub : sub_client.o client_shared.o ../lib/libmosquitto.a
-	${CROSS_COMPILE}${CC} $^ -o mosquitto_sub ${CLIENT_LDFLAGS} -lssl -lcrypto -lpthread
+	${CROSS_COMPILE}${CC} --sysroot=${SYSROOT} $^ -o mosquitto_sub ${CLIENT_LDFLAGS} -lssl -lcrypto -lpthread

Am I missing something? Maybe is CFLAGS passed every time that you run ${CC}, my understanding is that unless you pass CFLAGS to GCC it will not use it. Even when most of make files use this construction, since it is encouraged in all manuals, I'm not able to see it in those Makefiles.

dezquerra:
@ralight  That's true, i was focusing on makefile and din't check for the config.mk.

PR #738, opened 2018-03-23
Add is session present

bettio:
Add getter for session present flag to the API.

### DIFF
--- a/lib/cpp/mosquittopp.cpp
+++ b/lib/cpp/mosquittopp.cpp
@@ -269,6 +269,11 @@ int mosquittopp::opts_set(enum mosq_opt_t option, void *value)
 	return mosquitto_opts_set(m_mosq, option, value);
 }
 
+bool mosquittopp::is_session_present_on_connect()
+{
+	return mosquitto_is_session_present(m_mosq);
+}
+
 int mosquittopp::threaded_set(bool threaded)
 {
 	return mosquitto_threaded_set(m_mosq, threaded);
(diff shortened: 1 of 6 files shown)


karlp:
What's the usecase for this?

bettio:
Session present flag is part of the MQTT 3.1.1 specification and mosquitto API already allows to request a persistent session, but the client needs to know if the present flag (that is sent from the server) has been set or not/if the persistent session is already there, so those changes allow to know that.

bettio:
Does anybody knows why the CI fails? I can't reproduce that issue here.

Tifaifai:
After merge of roger and connack story ;)
```
./08-tls-psk-pub.py
FAIL: Received incorrect connack.
Received: CONNACK, rl=2, res=0, rc=5
Expected: CONNACK, rl=2, res=0, rc=0
...
Traceback (most recent call last):
  File "./08-tls-psk-pub.py", line 47, in <module>
    sock = mosq_test.do_client_connect(connect_packet, connack_packet, port=1889, timeout=20)
  File "/home/travis/build/eclipse/mosquitto/test/mosq_test.py", line 75, in do_client_connect
    raise ValueError
ValueError
```

bettio:
Any news here?

ralight:
Thanks very much for this contribution, but I've decided to implement it in a different way. See 86ced4d98fbe244c35ccc6a248862f06a2089fd4 and ff79432364fd87827c81838ba7ac3e7e6dc780a3. Does that seem reasonable to you?

For interest, the problem with the test is that when libmosquitto is compiled, a linker script ensures that only certain functions are exposed. You'd not added your function to the linker script, so the C++ library couldn't see that function. If you were compiling on Windows/Mac then you wouldn't have seen this, it only affects *nix.

bettio:
Thanks a lot, I'll close this pull request.

PR #749, opened 2018-03-29
remove dead code

hailinzeng:
inside the `if(!temp->payload)` branch, the statement `if(temp->payload) _mosquitto_free(temp->payload);` is dead code.


ralight:
Yes, that's pretty clear. Thanks very much for picking it up.

PR #750, opened 2018-03-29
Fix SSL on Windows

t-beckmann:
When trying to connect to MQTT over SSL on Windows the broker immediately disconnects the client.

Debugging shows the issue is that _mosquitto_net_read sets the errno variable to EAGAIN, but the code at the calling side in _mosquitto_packet_read overwrites the value assigning WSAGetLastError() on Windows platform. The WSAGetLastError() indicates no failure, but the following code does not expect that value.

The fix is to put back the altered errno using WSASetLastError() so that it can be picked up again.

### DIFF
--- a/lib/net_mosq.c
+++ b/lib/net_mosq.c
@@ -784,6 +784,9 @@ ssize_t _mosquitto_net_read(struct mosquitto *mosq, void *buf, size_t count)
 				}
 				errno = EPROTO;
 			}
+#ifdef WIN32
+			WSASetLastError(errno);
+#endif
 		}
 		return (ssize_t )ret;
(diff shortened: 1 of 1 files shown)


t-beckmann:
By the way, I'm using Windows 10 and OpenSSL 1.1.0 and compiled stuff from source WITH_TLS.

ralight:
Thank you, I like easy fixes with good explanations. I've applied this to the develop branch instead, so I'm going to close this PR. It will be released soon.

ralight:
Change of plan, it can go on the fixes branch too.

ralight:
Thanks for your contribution!

Issue #756, opened 2018-04-01
A bridge waits for restart timeout to connect to broker when using ADNS


lwindg:
When using ADNS, the first time to check status is **always** `EAI_INPROGRESS` (from my test); that cause the bridge spends at least "**restart timeout**" (default 30 seconds) to connect to the broker.

I found `restart_t` is set to `1` for quick restart ([bridge.c#L135](https://github.com/eclipse/mosquitto/blob/master/src/bridge.c#L135)), but it seems not enough. 

Maybe it should allow more time for DNS lookup.

ralight:
I think this has been fixed in 1.5.3 and the upcoming 1.5.4. I'm going to close this, if it's still an issue please reopen.

PR #757, opened 2018-04-01
Fix for waiting restart timeout with ADNS (#756)

lwindg:
The first time to check DNS resolve status may be `EAI_INPROGRESS` in most cases. 
This fix take `restart_t == 1` as a flag that should quickly restart for DNS lookup and will try 10 times before reset to "restart timeout".

I don't think using `restart_t` is a good idea, maybe be a `status` or `state` is better.

Signed-off-by: Yingyou Chen <lwindg@gmail.com>


imZack:
@ralight Are you able to review this one?  Thanks

PR #766, opened 2018-04-05
Fix: use websockets without OpenSSL / TLS

Tifaifai:
If you use websockets without OpenSSL / TLS support, you don't compile mosquitto.

THX


Tifaifai:
Problem with signed-off

PR #768, opened 2018-04-06
Fix: use websockets without OpenSSL / TLS.

Tifaifai:
If you use websockets without OpenSSL / TLS support, you don't compile mosquitto.

THX
Signed-off-by: Tifaifai Maupiti <tifaifai.maupiti@gmail.com>


ralight:
Looks good to me, thanks very much!

PR #771, opened 2018-04-07
Resolved redundant listener lookup in socket accept on TLS init.

qubeck:


ralight:
This has now been superseded by other changes. Thanks very much for spending the time on it though.

Issue #706, opened 2018-02-26
http: serving "abc/" as "abc/index.html"


jcw:
Mosquitto could be the single-tool solution for (quasi-) serverless applications, i.e. rich JavaScript applications (React, Angular, etc) which load from a 100% static web server, and then do all their dynamic work through WebSockets and MQTT. I think there's just one _tiny_ enhancement missing.

Right now, if I understand correctly, Mosquito can serve through a websocket, and it can server static files over http. Unfortunately, the default mode for React apps (and tools such as Gatsby) is to use sanitised URLS, i.e. "http://.../abc/" instead of "http://.../abc/index.html" or ""http://.../abc.html".

This leads to errors here: https://github.com/eclipse/mosquitto/blob/master/src/websockets.c#L519

Would it be possible to add logic to look for the `index.html` file if a directory path is encountered? I can probably create a pull request for such an enhancement, but my question is really: would there be any interest to add something like this to the official repo, if the change is reviewed and found to properly address this issue?

ralight:
Yes, that sounds like a sensible suggestion, thanks for bringing it up.

ralight:
I've just pushed a commit to the develop branch that should fix this, would you be able to test it out?

PR #791, opened 2018-04-20
 Fix: To correct #706 with Linux/Windows support stat.h

Tifaifai:
In the commit following issue #706, filestat.st_mode protection was missing for Linux or Windows compilation support. We also cast the use of realpath function to prevent two warnings.


Tifaifai:
As you want but we have need this patch for compilation on RPi. We have tested on Raspberry Pi3 / Arm Arch. You can compare with this link https://github.com/eclipse/mosquitto/commit/9ea0d38002da0c2f0956243c21eb7dfd6e5d490e#diff-849b82dfc54d09a3b7c5dca6778c79c8L520

Maybe a problem with our compilation toolchain but we don't access in /usr/include/arm-linux-gnueabihf/bits/stat.h directory. This patch resolv our problem.

karlp:
I'm with ral, this looks like you're toolchain is busted or not installed properly.

ralight:
If this fixes the build for you, could you explain what error you actually get without it? I have no problems building on a Raspberry Pi 1.

Tifaifai:
With Raspberry Pi3 / Latest Raspbian Light and **Without this patch** : 
```
pi@Tifaifai-pi2:~/i2c $ make -j 4
net_mosq.c:446:12: warning: ‘net__init_ssl_ctx’ defined but not used [-Wunused-function]
 static int net__init_ssl_ctx(struct mosquitto *mosq)
            ^~~~~~~~~~~~~~~~~
../lib/net_mosq.c:446:12: warning: ‘net__init_ssl_ctx’ defined but not used [-Wunused-function]
 static int net__init_ssl_ctx(struct mosquitto *mosq)
            ^~~~~~~~~~~~~~~~~
security_default.c:709:12: warning: ‘mosquitto__memcmp_const’ defined but not used [-Wunused-function]
 static int mosquitto__memcmp_const(const void *a, const void *b, size_t len)
            ^~~~~~~~~~~~~~~~~~~~~~~
websockets.c: In function ‘http__canonical_filename’:
websockets.c:444:23: warning: implicit declaration of function ‘realpath’ [-Wimplicit-function-declaration]
  filename_canonical = realpath(filename, NULL);
                       ^~~~~~~~
websockets.c:444:21: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
  filename_canonical = realpath(filename, NULL);
                     ^
websockets.c: In function ‘callback_http’:
websockets.c:531:13: warning: implicit declaration of function ‘fileno’ [-Wimplicit-function-declaration]
    if(fstat(fileno(u->fptr), &filestat) < 0){
             ^~~~~~

websockets.c:540:27: error: ‘S_IFDIR’ undeclared (first use in this function)
    if((filestat.st_mode & S_IFDIR) == S_IFDIR){
                           ^~~~~~~
websockets.c:540:27: note: each undeclared identifier is reported only once for each function it appears in
websockets.c:552:27: error: ‘S_IFREG’ undeclared (first use in this function)
    if((filestat.st_mode & S_IFREG) != S_IFREG){
                           ^~~~~~~

websockets.c: In function ‘mosq_websockets_init’:
websockets.c:703:18: warning: assignment makes pointer from integer without a cast [-Wint-conversion]
   user->http_dir = realpath(listener->http_dir, NULL);
                  ^
make[1]: *** [websockets.o] Erreur 1
```

karlp:
is this perhaps jsut something that isn't being included because it often gets included implicitily via another header?  are you building mosquitto with ssl but websockets without?  what toolchain?

Tifaifai:
For this test, mosquitto is compile without ssl and libwebsockets without ssl.
Compilation on Raspberry Pi3 with classic tools (libc6-dev, g++, cmake, uuid-dev, build-essential)

ralight:
I believe this should be fixed for you now. Could you please confirm on the latest develop branch changes.

Tifaifai:
Solution : you can replace _BSD_SOURCE by _GNU_SOURCE.

From the Linux man page on [feature test macros](http://man7.org/linux/man-pages/man7/feature_test_macros.7.html):
```
Since glibc 2.19, defining _GNU_SOURCE also has the effect of implicitly defining
_DEFAULT_SOURCE. In glibc versions before 2.20, defining _GNU_SOURCE also 
had the effect of implicitly defining _BSD_SOURCE and _SVID_SOURCE.
```
It's OK for me ;)

Issue #763, opened 2018-04-03
Publish actual version to docker hub


nixikanius:
As I see in [releases page](https://github.com/eclipse/mosquitto/releases) current actual version of the mosquitto is **1.4.15**. But the latest published version of the mosquitto at [docker hub](https://hub.docker.com/r/library/eclipse-mosquitto/tags/) is **1.4.12**, which is almost 1 year old.

I have one problem with persistence functionality and want to check the mosquitto at last version using docker. Please push the actual version of the mosquitto to docker hub.

Thanks.

chriswue:
Looks like 1.5 should be on there soon as well: https://github.com/docker-library/official-images/pull/4987

ralight:
1.5.3 is indeed now on the official image and should be kept up to date now, so I'm closing this.

PR #793, opened 2018-04-22
add dockerfile for 1.4.15 (GH #763)

winklerand:
Add a Dockerfile for mosquitto 1.4.15

partially addresses issue #763

Signed-off-by: Andreas Winkler <winkler.andreas2@gmail.com>

### DIFF
--- a/docker/1.4.15/Dockerfile
+++ b/docker/1.4.15/Dockerfile
@@ -1,0 +1,13 @@
+FROM alpine:3.7
+LABEL maintainer="David Audet <david.audet@ca.com>"
+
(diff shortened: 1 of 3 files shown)


efrecon:
I think that #799 solves the problem more elegantly as it could generate Docker images for each official release of mosquitto independently of when these are taken into Alpine as a repository. This would provide the project complete control of the Docker image instead of relying on an external packaging party (Alpine).

winklerand:
I agree with @efrecon, closing this PR in favor of #799 which is the better approach IMHO.

Issue #669, opened 2018-01-09
mosquitto 1.4.14 on official docker repository


Foxhunt:
Hi,

is it possible to publish mosquitto 1.4.14 on the official eclipse-mosquitto docker repository?

It has been six months since it got added to the develop branch with #500.
This was already addressed in #606, but not commented by any maintainer.

Is there a place to get in touch? The Paho channel on the clipse.org Mattermost server is kind of dead. So i'm asking here.

vinceskahan:
Doesn't it depend on what packages Alpine Linux provides upstream ?  Read the Dockerfile.

Edit the Dockerfile to use alpine:3.7 and specify mosquitto=1.4.14-r3 in it and build your own container that way.  Works fine.

Foxhunt:
That is what we ended up doing.

`FROM eclipse-mosquitto:latest ` should provide you the _latest_ mosquitto version, but it doesen't.
The needed Dockerfile is already on the develop branch, so it could be published to the official Docker Hub repository. So building your own alpine + mosquitto image would not be necessary.

vielmetti:
v1.5 is out now, and I've noted this at https://github.com/docker-library/official-images/issues/4358

J0WI:
https://github.com/docker-library/official-images/blob/master/library/eclipse-mosquitto should be updated

sasincj:
Hi, Why docker hub is not updated with latest mosquitto version. When mosquitto 1.5.3 will be available at docker hub?

ralight:
1.5.3 is now available on the official image and will be kept up to date.

Issue #606, opened 2017-11-04
Docker image availability


ruimarinho:
Hi,

The latest official image available on [Docker Hub](https://hub.docker.com/_/eclipse-mosquitto/) contains a vulnerability fixed in mosquitto v1.4.12+.

I can see that the master branch [does not contain versions](https://github.com/eclipse/mosquitto/tree/master/docker) 1.4.10+, while the [developer branch does](https://github.com/eclipse/mosquitto/tree/develop/docker). Since v1.4.14 has been released, is there any reason for not having this merged into master?

cc @daudetCA

lukecyca:
I am using 1.4.12 because that's the latest available as an official mosquitto image on Docker Hub.

It's disappointing that official docker image releases don't coincide with official mosquitto releases.I would expect that 1.4.14 be on Docker Hub.

ruimarinho:
In the meantime I've created https://hub.docker.com/r/ruimarinho/mosquitto/.

lukecyca:
Thanks. I'll use yours for now then.

triplepoint:
I, too,  would appreciate keeping the Docker Hub mosquitto image up to date with project releases.

You can configure github to notify Docker Hub when new git tags are pushed; this would be the most reliable and hands-free way to do it.

https://docs.docker.com/docker-hub/builds/#link-to-a-hosted-repository-service

Please take the time to set this up.

triplepoint:
Any news on this?  The current published docker container is still standing at 1.4.12, and I see 1.5.1 was released a week or so ago.  

Bugfixes aren't effective if they're not deployed.

efrecon:
Not that I know, I am also waiting for the maintainers to merge this and adapt the Docker image generation. I would be happy to help if necessary or if I've missed anything in the process.

triplepoint:
Ok, here's a PR adding a dockerfile that pulls the tarball specified in the build-arg.
https://github.com/eclipse/mosquitto/pull/926

The idea is this:
```
cd docker/from_dist
docker build -t eclipse-mosquitto:1.5.1 --build-arg VERSION="1.5.1" .
docker run --rm -it eclipse-mosquitto:1.5.1
```

So it's on the user to keep the docker image label matching the build-arg that was used when it was built, but otherwise it's as clear as translation of the Alpine package build script as I could make:
https://github.com/alpinelinux/aports/blob/master/main/mosquitto/APKBUILD

[...]

Issue #364, opened 2017-01-30
Can the clients be added to the Docker image?


jdiamond:
Most of the official Docker images containing servers (postgres, mysql, redis, mongo, etc) have the CLI tools installed in them.

It looks like it'd only require adding just one line to the Dockerfile. I can send a PR if you like.

ralight:
Yes please, just open a PR.

jdiamond:
I'm not signing the CLA so somebody else should do it.

ralight:
Just going through old issues, the clients are part of the docker image so I'm closing this.

PR #799, opened 2018-04-25
Adding compiled Alpine Docker image, generic for all versions

efrecon:
This makes it possible to generate Docker images for all releases of mosquitto, using the github tags that are associated to each release. Previous Docker images would rely on the packages that were present as part of the Alpine distribution. This implementation will, instead, compile from source, thus removing the dependency on Alpine (re)packaging of the project.

Generation of releases is by way of Docker support for build arguments. As a result, this Docker image generation implementation would provide the base for squeezing issues #606 #669. In addition, this image also includes the pub and sub clients, thus squeezing #364. Care has been taken to remain compatible with the previous implementations, through keeping the same directory structure for hosting the logs, configuration files, or persistence data.

As commented [here](https://github.com/eclipse/mosquitto/issues/606#issuecomment-357476814) it is possible to let github notify the Docker hub about new tags. This code would then be able to be used for generating the image based on the tag. 

Signed-off-by: Emmanuel Frécon <efrecon@gmail.com>

### DIFF
--- a/docker/generic/Dockerfile
+++ b/docker/generic/Dockerfile
@@ -3,40 +3,61 @@ MAINTAINER Emmanuel Frecon <efrecon@gmail.com>
 
 LABEL Description="Eclipse Mosquitto MQTT Broker"
 
-ARG VERSION=1.4.15
+ARG VERSION=1.5
+ARG WS_VERSION=3.0.0
 
 RUN set -ex \
     # Add build dependencies, remove after build
     && apk --no-cache add --virtual .build-deps \
         build-base \
-        libressl-dev \
+        openssl-dev \
         c-ares-dev \
         util-linux-dev \
-        libwebsockets-dev \
+        zlib-dev \
+        cmake \
     # Add fetch dependencies, remove after build
     && apk --no-cache add --virtual .fetch-deps \
         git \
         curl \
     # Add run dependencies, keep after build
     && apk --no-cache add --virtual .run-deps \
-        libressl \
+        openssl \
+        zlib \
         c-ares \
         util-linux \
-        libwebsockets \
+    # Manually compile and install libwebsockets to keep relying on openssl
+    && mkdir /tmp/libwebsockets \
+    && cd /tmp/libwebsockets \
+    && curl -sqL https://github.com/warmcat/libwebsockets/archive/v${WS_VERSION}.tar.gz | tar zxf - \
+    && cd libwebsockets-${WS_VERSION} \
+    && mkdir build \
+    && cd build \
+    && cmake .. \
+		-DCMAKE_BUILD_TYPE=MinSizeRel \
+		-DCMAKE_INSTALL_PREFIX=/usr \
+		-DCMAKE_VERBOSE_MAKEFILE=TRUE \
+    && make \
+    # Do not make install blindly as installing the examples is neither needed
+    # nor works.
+    && cmake -DCOMPONENT=libraries -P cmake_install.cmake \
+    && cmake -DCOMPONENT=dev -P cmake_install.cmake \
     # Checkout and compile, versions are branched with a "v" in front of the
     # version number, but tagged at the Docker registry without the "v".
     && git clone --depth 1 -b v${VERSION} https://github.com/eclipse/mosquitto.git /tmp/mosquitto \
     && cd /tmp/mosquitto \
-    && curl -qsL https://git.alpinelinux.org/cgit/aports/plain/main/mosquitto/libressl.patch -o libressl.patch \
-    && git apply libressl.patch \
+    # Manually patch the Makefile to arrange for not making the documentation
+    # (and not having to bring in xslt). This is to be able to run against older
+    # version of Mosquitto where WITH_DOCS did not (yet) exist.
+    && sed -i -e "s|^DOCDIRS=.*$|DOCDIRS=|g" Makefile \
+    # Since we compile under Alpine, with musl, we cannot add support for async
+    # dns lookup as this requires glibc
     && make \
-        WITH_MEMORY_TRACKING=no \
         WITH_WEBSOCKETS=yes \
         WITH_SRV=yes \
-        WITH_TLS_PSK=no \
         WITH_ADNS=no \
-        WITH_DOCS=no \
         prefix=/usr \
+    # Manually patch the Makefile to make sure we can still run on older
+    # releases of Mosquitto where WITH_STRIP did not (yet) exist.
     && sed -i -e "s|(INSTALL) -s|(INSTALL)|g" -e 's|--strip-program=${CROSS_COMPILE}${STRIP}||' */Makefile */*/Makefile \
     && make WITH_DOCS=no prefix=/usr install \
     && addgroup -S mosquitto \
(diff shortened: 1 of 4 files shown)


docker/generic/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+        util-linux \
+        libwebsockets \
+    # Checkout and compile, versions are branched with a "v" in front of the
+    # version number, but tagged at the Docker registry without the "v".
+    && git clone --depth 1 -b v${VERSION} https://github.com/eclipse/mosquitto.git /tmp/mosquitto \
+    && cd /tmp/mosquitto \
+    && curl -qsL https://git.alpinelinux.org/cgit/aports/plain/main/mosquitto/libressl.patch -o libressl.patch \
+    && git apply libressl.patch \
+    && make \
+        WITH_MEMORY_TRACKING=no \
```
karlp: why are you turning this off?

docker/generic/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+        WITH_WEBSOCKETS=yes \
+        WITH_SRV=yes \
+        WITH_TLS_PSK=no \
```
karlp: why are you turning this off?

docker/generic/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+        WITH_ADNS=no \
+        WITH_DOCS=no \
+        prefix=/usr \
+    && sed -i -e "s|(INSTALL) -s|(INSTALL)|g" -e 's|--strip-program=${CROSS_COMPILE}${STRIP}||' */Makefile */*/Makefile \
```
karlp: what's the purpose of this line? Is this really indicating a makefile deficiency that should be addressed separately?

docker/generic/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+    && git clone --depth 1 -b v${VERSION} https://github.com/eclipse/mosquitto.git /tmp/mosquitto \
+    && cd /tmp/mosquitto \
+    && curl -qsL https://git.alpinelinux.org/cgit/aports/plain/main/mosquitto/libressl.patch -o libressl.patch \
+    && git apply libressl.patch \
+    && make \
+        WITH_MEMORY_TRACKING=no \
+        WITH_WEBSOCKETS=yes \
+        WITH_SRV=yes \
+        WITH_TLS_PSK=no \
+        WITH_ADNS=no \
```
karlp: Why are you turning this off?

docker/generic/Dockerfile
efrecon: This is taken from what the Alpine people were doing when compiling mosquitto.

docker/generic/Dockerfile
efrecon: Same as above. I tried achieving maximum compatibility with the Alpine-based (and packaged) version of mosquitto.

docker/generic/Dockerfile
efrecon: Yes, at least, I thought it was, and I now see that it has been fixed by d90cd585dd8cc1f9cecd9082d71333e1e638df2d. I hadn't seen that before you pointed it out. Again, this behaviour is inherited from the way Alpine does it when [compiling](https://git.alpinelinux.org/cgit/aports/plain/main/mosquitto/APKBUILD?h=3.4-stable) mosquitto.

docker/generic/Dockerfile
efrecon: Same again, for maximum compatibility with how the Alpine package was [compiled](https://git.alpinelinux.org/cgit/aports/tree/main/mosquitto/config.patch?id=f61baacc21feda8c0f7177144a4dd3ee81aa7735).

efrecon:
I focused on making the transition from the old way (Alpine package) to the new way (home-grown) Docker image as smooth as possible for users. This might not make sense and the project would perhaps want for some of the `WITH_` variables not to be turned off. Maybe is the proper way to progress to first do as I tried and later introduce back some of those features as part of the Docker image?

docker/generic/Dockerfile
karlp: Perhaps alpine is making poor choices?  This is somethign that used to be off in some distributions because uclibc didn't support it.  I believe musl does?  You seem to be simply cargo culting settings with no real basis for them though.

docker/generic/Dockerfile
karlp: So, turn it back on, and if it doens't work, complain to alpine that their ssl implementation is broken and incomplete.  This is definitely functionality that is in use.

docker/generic/Dockerfile
karlp: Yes, copying things blindly from some random downstream distro is not necessarily a path to glory.

docker/generic/Dockerfile
efrecon: You *do* understand that this is what the *official* (at least tagged as so, seen from most Docker users) image for [mosquitto](https://hub.docker.com/_/eclipse-mosquitto/) was compiled? This state of affairs was the reason why I did not want to depart from the current state. As I pointed out in my comment further down, this would mean that the project makes a decision about what the Docker image would contain, which is fine and which we can do here. Let me try this and the other switches that you pointed out and return to you.

efrecon:
I have just pushed a number of changes to the branch. This is the result of a number of hours work to try turning on as many of the configuration flags as possible. In order to do this, I have:
* Brought in OpenSSL instead of LibreSSL, which is otherwise the preferred SSL implementation on Alpine.
* Arranged to fetch, release and install [libwebsockets](https://github.com/warmcat/libwebsockets/) since it also depends on an SSL implementation.
* Made the version of libwebsockets a Docker build argument.

These changes lead to having `WITH_MEMORY_TRACKING` and `WITH_TLS_PSK` turned on, via their default values in `config.mk`. However, I have not been able to:
* Turn on asynchronous DNS lookup, as the comment in `config.mk` mentions a strict dependency on `glibc` for being able to turn on `WITH_ADNS`. Alpine is running with musl.
* I did not use `WITH_STRIP=no`, since this would otherwise prevent this Dockerfile to generate older versions of mosquitto. The feature was added a while ago.
* For the same reason, I actually stopped using `WITH_DOCS=no` and opted to manually patched the Makefile. This allows the Dockerfile to be able to compile older releases (tried on [1.4.8](https://github.com/eclipse/mosquitto/tree/v1.4.8)).

Remember, by default this Dockerfile compiles the latest tagged version of mosquitto, i.e. `1.5` at the time of writing. However, the image is able to compile most of the past releases of the project, based on their respective tags in the repo. It does so by cloning the source tree for the given tag and compiling it, and uses a similar technique for libwebsockets. The mosquitto and libwebsockets version numbers are controlled by Docker build arguments.

The implementation notes provide a bit more details and comments in the Dockerfile provide the rationale behind the decisions taken around `strip` and the documentation. I hope that you will now be able to accept this PR.

martinvw:
It would be great if this or a similar PR could be merged, its a pity that there are no recent docker builds.

ralight:
Thank you for taking the time to write this update. I have produced my own dockerfile that compiles from the published tarballs. I hope you don't mind me not using your work, but I felt it necessary to fully understand the work I was committing by spending time rewriting to get it into the state it is now. I'm now in a much better position to keep things up to date.

As similar work is now in the repo, I'm going to close this but thanks again for spending time on it.

efrecon:
I had a quick look at the new Dockerfile. I think that this is fine. The most important thing is that the project gets updated images of all 1.5.X and future versions of mosquitto, without lagging behind has much has it has had in the past. Great work!

Issue #796, opened 2018-04-24
Cannot configure acl_file on a per-listener basis


jpflorijn:
According to the man page:
```
       per_listener_settings [ true | false ]
           If true, then authentication and access control settings will be controlled on a per-listener basis. The following
           options are affected:

           password_file, acl_file, psk_file, allow_anonymous, allow_zero_length_clientid, auth_plugin, auth_opt_*,
           auto_id_prefix.
```

So I wrote the following mosquitto.conf:
```
per_listener_settings true

listener 1883 localhost
acl_file first.acl

listener 1884 localhost
acl_file second.acl
```
and ran mosquitto against it, with the following result:
```
Error: Duplicate acl_file value in configuration.
Error found at mosquitto.conf:7.
Error: Unable to open configuration file.
```

I am running the current HEAD of the develop branch (`f2376551bc0a91368df3d3784aea2e3a3bf68bbf`).

Looking at `src/conf.c` I noticed the `acl_file` is not part of the per-listener settings. Is this an oversight? Or is it still to come?

jpflorijn:
I've contributed a patch for this issue. Let me know if it needs anything else.

ralight:
@jpflorijn Thanks for pointing this out, it was an oversight in as much as I'd done most of the work but not completed it or committed it. I've submitted a change that implements this. I really do appreciate the time you've taken on your PR, I haven't used it though because I'd already done a good deal of the work and for a very practical reason - it is problematic to take contributions at this point because the IP log of external contributions for this release has already been approved.

jpflorijn:
@ralight So long as it gets fixed I'm happy. Thanks for looking into it!

PR #801, opened 2018-04-25
Implemented per-listener acl_files, as per documentation. Fixes eclipse/mosquitto#796.

jpflorijn:
I thought I'd fix the issue I encountered. I hope I worked in a way that is acceptable to you.
I was not entirely sure of the modified function signature of `aclfile__parse(...)` with the double arguments, but declaring a union or putting the acl_list and acl_patterns in yet another struct also felt clunky to me and would've caused even more changes to the codebase.

I have tested this with per-listener acl_files and it seems to work, also across reloads. Not sure what else to test.

If there is anything else I need to do, let me know.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -146,8 +146,8 @@ static int conf__attempt_resolve(const char *host, const char *text, int log, co
 static void config__init_reload(struct mosquitto__config *config)
 {
 	/* Set defaults */
-	mosquitto__free(config->acl_file);
-	config->acl_file = NULL;
+	mosquitto__free(config->security_options.acl_file);
+	config->security_options.acl_file = NULL;
 	config->security_options.allow_anonymous = -1;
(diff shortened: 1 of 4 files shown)


ralight:
As I've mentioned in your bug report, there are practical reasons why it's awkward to take this code for the 1.5 release, but in addition to that you can see the ip-validation step has failed because you haven't signed the Eclipse Contributor Agreement. I'd be very happy to take a PR from you in the future, but you will need to sign the ECA to have them accepted.

jpflorijn:
I did sign the ECA though, but only after I opened the PR. It probably doesn't re-do ip-validation step then, even though I forced another push to the branch. The issue is resolved so I'm happy.

ralight:
Ok, sorry I didn't check - I would do normally when looking to accept a PR. I just wanted to make sure you were aware.

PR #807, opened 2018-05-02
Fix: Compile without OpenSSL / TLS.

Tifaifai:
Same problem that #768, If you compile without OpenSSL / TLS support, you don't compile mosquitto.

THX


Tifaifai:
Sorry for Commit Summary ;) changed the base branch !

ralight:
Thanks for your contribution - as you can see the ip-validation step failed so I' afraid cannot accept it as it is. It was easy to fix just from the title of the bug, so I've snuck my own fix in.

PR #808, opened 2018-05-03
Fix: Compile without OpenSSL / TLS.

Tifaifai:
Identical to (Fix: Compile without OpenSSL / TLS. #807).
Re-tested without the ip-validation step failed.


Tifaifai:
I don't understand why i have the ip-validation failed ! maybe the previous merge !
I'm reclose

PR #828, opened 2018-05-22
add missing paramters to internal mosquitto_acl_check

Daedaluz:
All fields in mosquitto_acl_msg wasn't filled before call to plugin.  
I'm not sure if there is another way this should be fixed, but i thought that i'd just add the missing struct fields through parameters to the mosquitto_acl_check call.

### DIFF
--- a/src/handle_subscribe.c
+++ b/src/handle_subscribe.c
@@ -112,7 +112,7 @@ int handle__subscribe(struct mosquitto_db *db, struct mosquitto *context)
 			log__printf(NULL, MOSQ_LOG_DEBUG, "\t%s (QoS %d)", sub, qos);
 
 			if(context->protocol == mosq_p_mqtt311){
-				rc = mosquitto_acl_check(db, context, sub, MOSQ_ACL_SUBSCRIBE);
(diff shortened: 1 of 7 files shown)


src/handle_subscribe.c
```diff
@@ -112,7 +112,7 @@ int handle__subscribe(struct mosquitto_db *db, struct mosquitto *context)
 			log__printf(NULL, MOSQ_LOG_DEBUG, "\t%s (QoS %d)", sub, qos);
 
 			if(context->protocol == mosq_p_mqtt311){
-				rc = mosquitto_acl_check(db, context, sub, MOSQ_ACL_SUBSCRIBE);
+				rc = mosquitto_acl_check(db, context, sub, 0, NULL, 0, false, MOSQ_ACL_SUBSCRIBE);
```
karlp: surely you would provide the requested qos here, not 0?

Daedaluz: Good point.. missed that. :)

ralight:
Lovely, thanks very much for this and sorry it has taken so long to merge. It will be in 1.5.1 which will probably be next week.

Issue #826, opened 2018-05-20
Undefined reference libwebsocket_context_destroy


sddan:
When building mosquitto 1.5 with websockets from source, I ran into the following error

> set -e; for d in lib client src; do make -C ${d}; done
> make[1]: Entering directory '/home/pi/Downloads/mosquitto-1.5/lib'
> make -C cpp
> make[2]: Entering directory '/home/pi/Downloads/mosquitto-1.5/lib/cpp'
> make[2]: Nothing to be done for 'all'.
> make[2]: Leaving directory '/home/pi/Downloads/mosquitto-1.5/lib/cpp'
> make[1]: Leaving directory '/home/pi/Downloads/mosquitto-1.5/lib'
> make[1]: Entering directory '/home/pi/Downloads/mosquitto-1.5/client'
> make[1]: Nothing to be done for 'all'.
> make[1]: Leaving directory '/home/pi/Downloads/mosquitto-1.5/client'
> make[1]: Entering directory '/home/pi/Downloads/mosquitto-1.5/src'
> cc -Wall -ggdb -O2  -I. -I.. -I../lib  -DVERSION="\"1.5\"" -DWITH_BROKER -DWITH_TLS -DWITH_TLS_PSK -DWITH_UUID -DWITH_BRIDGE -DWITH_PERSISTENCE -DWITH_MEMORY_TRACKING -DWITH_SYS_TREE -DWITH_WEBSOCKETS -DWITH_EC -DWITH_EPOLL -c loop.c -o loop.o
> cc -Wall -ggdb -O2  -I. -I.. -I../lib  -DVERSION="\"1.5\"" -DWITH_BROKER -DWITH_TLS -DWITH_TLS_PSK -DWITH_UUID -DWITH_BRIDGE -DWITH_PERSISTENCE -DWITH_MEMORY_TRACKING -DWITH_SYS_TREE -DWITH_WEBSOCKETS -DWITH_EC -DWITH_EPOLL -c ../lib/memory_mosq.c -o memory_mosq.o
[...]

PR #830, opened 2018-05-23
Recent version libwebsockets

Tifaifai:
Hi,
In relation with 'Undefined reference libwebsocket_context_destroy' #826, 
i propose you to upgrade the minimum version of libwebsockets of 1.3 to 1.6

For example, the last update on 1.5-stable is 21 Jan 2016 and on 1.6-stable is 22 Apr 2016.

The v2.4-stable branch is always active and the last release (3.0-stable) is 5 May 2018.

C-U

### DIFF
--- a/compiling.txt
+++ b/compiling.txt
@@ -1,12 +1,11 @@
 The following packages can be used to add features to mosquitto. All of them
 are optional.
 
-<<<<<<< HEAD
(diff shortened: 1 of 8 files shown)


ralight:
I shouldn't have changed this to fixes, never mind though. I'm not changing what versions we support in a bugfix release, but I'll be revisiting this for the next feature release - there is a lot of cruft related to old versions of LWS.

PR #835, opened 2018-05-26
Add documentation and networking dep to unit files

richmattes:
Updated the unit file examples to add a dependency on the
network-online.target systemd target to prevent mosquitto from
starting without network connectivity.

Added a documentation line to the unit files pointing to mosquitto
manpages.

Signed-off-by: Rich Mattes <richmattes@gmail.com>


ralight:
Thanks very much, and sorry for the slow response.

PR #839, opened 2018-05-28
add a maximum limit to the output queue 

chusAlvarez:
Currently the output queue is not limited, only the number of inflight messages are.
This change add a control in the queue grow, to prevent it from growing unontrollably when the connection to the brooker is lost.

A 0 value means "unlimited"
Default value is 0, for retro-compatibility with older code

When the limit is reached, the packages are dropped and mosquitto_publish returns a new error code, MOSQ_ERR_QUEUE_LIMIT

Signed-off-by: chus alvarez <chus.alvarez@fon.com>

### DIFF
--- a/lib/linker.version
+++ b/lib/linker.version
@@ -90,4 +90,5 @@ MOSQ_1.5 {
 		mosquitto_sub_topic_check2;
 		mosquitto_topic_matches_sub2;
 		mosquitto_connect_with_flags_callback_set;
+		mosquitto_max_pub_queue_messages_set;
 } MOSQ_1.4;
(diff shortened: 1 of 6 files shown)


lib/linker.version
```diff
@@ -90,4 +90,5 @@ MOSQ_1.5 {
 		mosquitto_sub_topic_check2;
 		mosquitto_topic_matches_sub2;
 		mosquitto_connect_with_flags_callback_set;
+		mosquitto_max_pub_queue_messages_set;
```
ralight: Thanks for spotting this important file. The new function will have to be in a new section, could you create it at the bottom of the file like:
```
MOSQ_1.6 {
    global:
        <function>
} MOSQ_1.5;
```

ralight (changes requested):
This generally looks good and useful. I've made a comment below for a change that needs to be made. I'm also not entirely sure about the new function name, `mosquitto_max_queued_messages_set` or `mosquitto_max_outgoing_messages_set` might work better, but they both still aren't quite right.
How  do you feel about writing a test for this as well?

PR #862, opened 2018-06-18
websockets: _GNU_SOURCE needed for S_IF{DIR,REG}

ffontaine:
Define of _GNU_SOURCE is needed to be able to use S_IFDIR and S_IFREG in
src/websockets.c

Fixes:
 - http://autobuild.buildroot.net/results/7dcfb6ca9d14a5cd6872590065549356f1ab42a0

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/websockets.c
+++ b/src/websockets.c
@@ -29,6 +29,8 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #ifdef WITH_WEBSOCKETS
 
+#define _GNU_SOURCE
+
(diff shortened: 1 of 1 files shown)


ralight:
What libc and version are you using here?

ralight:
This should now be fixed for version 1.5.4, but in a slightly different way (the define going in config.h), so I'm going to close this. Thanks for reporting the issue!

PR #864, opened 2018-06-21
Fixes #860

### LINKED ISSUE #860: Publish denied on bridge connection when per_listener_settings is true
My configuration is 

```
pid_file /var/run/mosquitto.pid
log_dest file /var/log/mosquitto/mosquitto.log

persistence true
persistence_location /var/lib/mosquitto/

per_listener_settings true


listener 1883 0.0.0.0
allow_anonymous true


listener 8883 0.0.0.0
password_file /etc/mosquitto/pwfile
allow_anonymous false
cafile /etc/mosquitto/ssl/ca/server-ca.crt
certfile /etc/mosquitto/ssl/internal-mqtt-broker.crt
keyfile /etc/mosquitto/ssl/internal-mqtt-broker.pem
tls_version tlsv1

connection to-external
address 172.16.1.10:8883
remote_password xxx
remote_username homegear
local_password xxx
local_username argus
bridge_insecure true
bridge_tls_version tlsv1
bridge_cafile /etc/mosquitto/ssl/ca/server-ca.crt
topic # both
```

Whenever a message from the other side of the bridge is received I see

```
1528996259: Denied PUBLISH from local.homegear.to-external (d0, q0, r0, m0, 'IoT/Test', ... (5 bytes))
```

in the logfile.

As soon as I remove the `per_listener_settings` entry everything is fine and the message is accepted.

Removing the `local_password` and `local_username` from the configuration doesn't change the behavior.

Is there anything wrong with my configuration - I don't think so - or is this a bug?

Cheers,
Wolfgang

wollud1969:
These changes fix the issue I've reported under the ID 860 in eclipse/mosquitto. (Bridges won't work anymore once the new per_listener_settings option is activated.)
I've provided a test, mostly stolen from one of the other bridge tests, just extended with an additional listener the the option per_listener_settings set to true, which is failing without the modification in security_default.c and which passes after applying this modification.
I've created my branch fix_860 on top of the master branch since the problem doesn't appear yet in the fixes branch (per_listener_settings are not yet available right there).
Eclipse CLA is signed, I would be happy to help with these changes.

### DIFF
--- a/src/security_default.c
+++ b/src/security_default.c
@@ -303,6 +303,7 @@ int mosquitto_acl_check_default(struct mosquitto_db *db, struct mosquitto *conte
 	struct mosquitto__security_options *security_opts = NULL;
 
 	if(!db || !context || !topic) return MOSQ_ERR_INVAL;
+	if(context->bridge) return MOSQ_ERR_SUCCESS;
 
 	if(db->config->per_listener_settings){
 		if(!context->listener) return MOSQ_ERR_ACL_DENIED;
@@ -314,7 +315,6 @@ int mosquitto_acl_check_default(struct mosquitto_db *db, struct mosquitto *conte
 			return MOSQ_ERR_PLUGIN_DEFER;
 	}
 
-	if(context->bridge) return MOSQ_ERR_SUCCESS;
 	if(access == MOSQ_ACL_SUBSCRIBE) return MOSQ_ERR_SUCCESS; /* FIXME - implement ACL subscription strings. */
 	if(!context->acl_list && !security_opts->acl_patterns) return MOSQ_ERR_ACL_DENIED;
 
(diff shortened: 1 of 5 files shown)


ralight:
I've merged this PR on the command line rather than here in github so that it could go on to the fixes branch instead. You can find the commits around here: https://github.com/eclipse/mosquitto/commit/88456c655f6a6deebf1a4c2bba13ec1ba2923f07

Thanks for taking the time to do this, it's especially nice when a good test gets included in a PR.

enen92:
I needed to backport this fix to an older version and had troubles finding the specific fix for https://github.com/eclipse/mosquitto/issues/860. Posting this comment so a cross reference between the fix and the issue is created.

enen92:
It'd be great if someone can cross-reference this PR on the original issue: https://github.com/eclipse/mosquitto/issues/860 .

PR #891, opened 2018-07-18
Flush stdout when asking for password.

ncopa:
Make sure the prompt is actaually printed by flushing stdout when asking
for passwords.


ralight:
This looks good, but in order to pass the IP check (which doesn't seem to have run) could you add a signed-off-by line to the commit? Rebasing it on the fixes branch would be helpful as well. I think you should just be able to do:

```
git commit --amend -s
git rebase origin/fixes
git push -f
```

ralight:
No need to change the base branch, I now know how to do it in the github interface.

ralight:
Thanks very much!

Issue #869, opened 2018-06-28
mosquitto_connect() fails on systems with no IPv4 addresses configured.


Alex-Richman:
mosquitto_connect() will fail on systems with no IPv4 addresses configured.  Even when connecting via the loopback interface to localhost.

e.g.:
```
ip addr flush dev eth0
./redacted 127.0.0.1
... [ Snip ] ...
Unable to connect: mosquitto_connect() failed (Lookup error)

ip addr add 192.168.0.1 dev eth0
./redacted 127.0.0.1
... [ Snip ] ...
Connected to broker on 127.0.0.1
```

From a brief look at the source, I believe this is because net__try_connect (lib/net_mosq.c:300) is using AI_ADDRCONFIG to enumerate local IP addresses to bind to.  Per the RFC this fails when the system has no IPv4 addresses configured (lo not being considered a configured address):
```
 - If the AI_ADDRCONFIG flag is specified, IPv4 addresses shall be
   returned only if an IPv4 address is configured on the local system,
   and IPv6 addresses shall be returned only if an IPv6 address is
   configured on the local system.  The loopback address is not
   considered for this case as valid as a configured address.
```

This may be related to https://github.com/eclipse/mosquitto/issues/736

Cheers,
- Alex.

lategoodbye:
Do we always need the flag AI_ADDRCONFIG for both getaddrinfo cases?

toast-uz:
I'd like to confirm the background. Disabling IPv4 addresses is frequently in these days? Many other network softwares have already corresponded with the situation?

Alex-Richman:
Certainly on embedded devices having no IPv4 addresses configured is very common (indeed having no interfaces except loopback is common).

In our case mosquitto is brokering messages between various modules which are themselves attempting to configure the system's IPv4 addresses as part of system startup (but are having a hard time since mosquitto needs the IPv4 addresses to be configured before it will connect...).

Either way I did patch this locally by removing the AI_ADDRCONFIG flag, which seems to work fine on our systems - I don't know if it's compatible with all configurations though.

Cheers,
- Alex.

toast-uz:
Thanks @Alex-Richman . It's great If you can submit your patch as a new PR.

I'd like to consult you. Should this issue be treated as a bug by lack of a primitive and nothing special feature, or as a new great and special feature?

[...]

PR #901, opened 2018-07-31
Remove AI_ADDRCONFIG flag from getaddrinfo() call.

Alex-Richman:
This avoids issues with systems where there are no interfaces configured (except loopback).  See #869.


karlp:
It's not when there's no ipv4 addresses configured, it's when there is nothing but loopback configured.

ralight:
Closed in a separate commit, thanks for your contribution anyway!

PR #927, opened 2018-08-23
fix SOCKS5 username\password set. fix SOCKS5 request

yzhivik:

-----

### DIFF
--- a/lib/socks_mosq.c
+++ b/lib/socks_mosq.c
@@ -80,7 +80,7 @@ int mosquitto_socks5_set(struct mosquitto *mosq, const char *host, int port, con
 		    	if(!mosq->socks5_password){
     			    	mosquitto__free(mosq->socks5_username);
         			return MOSQ_ERR_NOMEM;
-      			}
+			}
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


ralight:
Hello, thank you for your contribution. It looks like this is a worthwhile thing to include, but needs some work first.

Please split the separate issues into separate commits.
Please fix the formatting style to match the existing code.
Please try to remove the duplicate code from the ipv4/ipv6/domain section where possible.
Please complete the list of checkboxes in the original PR comment.

yzhivik:
@ralight Hello, fixed, please review

ralight:
I've ended up squashing your work into a single commit to avoid the reformatting fixes (for future reference, it's better to replace the original commits in your PR than add new ones on the end). This is now merged through the command line, so I'm going to close this. Thanks very much for your work!

yzhivik:
@ralight 
Thanks for pulling in!
I have local changes for:
1) PKCS12 container grinder (keys, certs and CAs)
2) Fix for TLS/SSL over proxy, lib tries to make secure connection with socks proxy instead of broker.
(This fix can be extended to distinguish or combine secured connection to broker or proxy or both)
WDYT?

PR #950, opened 2018-09-04
Install pkg-config files also for non-CMake builds

mhei:
At the moment, pkg-config hint files are only installed when CMake is
used as build system. However, it is very convenient for programs using
libmosquitto to have these files always in place, so let's add it
here, too.

-----

From my point of view, this is a fix - YMMV, so I could rebase to develop if desired...


karlp:
travis failure is about ssl cert validation failures, seems invalid for this PR? 
ssl.SSLError: [Errno 1] _ssl.c:510: error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed

PR #982, opened 2018-10-07
Fix mosquitto_pub -l if compiled using cmake.

majekw:
Since dde005ef92190dffd1bab0ddbfcfd8c5dbe17bc8 mosquito_pub is throwing error
that 'threading support has not been compiled' when compiled using cmake.
It looks like WITH_THREADING flag is not set at top level Makefile and used
only in lib/ directory, so library is correctly compiled with threading.
But for client this flag is undefined, so it gives error on '-l' option.

This commit is mostly cut&paste of part lib/CMakeLists.txt to correctly
define WITH_THREADING for client/ directory.

- [NA] If you are contributing a new feature, is your work based off the develop branch?
- [NA] Have you successfully run `make test` with your changes locally? (make test doesn't work with cmake)

-----


ralight:
Thanks for taking the time to submit this PR. I think it would be better not to duplicate the options, so would prefer to remove the part you have duplicated from lib/CMakeLists.txt and put it in the CMakeLists.txt in the root directory. That way both the lib and the client directories get the config from one place. If you want to keep the credit for the fix then please rework this PR, otherwise I am happy to do it myself.

If you do make the changes, could you also add an entry to the ChangeLog.txt in the 1.5.4 section:

```
Build:
- Fix clients not being compiled with threading support when using CMake. Closes #983.
```

I'd suggest editing the files as needed, then doing the following, which saves the need to open a new pull request:

```
git add CMakeLists.txt lib/CMakeLists.txt client/CMakeLists.txt ChangeLog.txt
git commit --amend
git push -f
```

majekw:
@ralight, thank you for suggestion and git commands. I moved whole section to top level CMakeLists.txt and it works also that way.

ralight:
Lovely, thanks very much :)

PR #1013, opened 2018-11-04
ASN1_STRING_data --> ASN1_STRING_get0_data

yoonghm:
ASN1_STRING_data() is similar to ASN1_STRING_get0_data() except the returned value is not constant. This function is deprecated: applications should use ASN1_STRING_get0_data() instead.

See https://www.openssl.org/docs/man1.1.0/crypto/ASN1_STRING_data.html

-----

### DIFF
--- a/src/handle_connect.c
+++ b/src/handle_connect.c
@@ -456,7 +456,7 @@ int handle__connect(struct mosquitto_db *db, struct mosquitto *context)
 						rc = 1;
 						goto handle_connect_error;
 					}
-					context->username = mosquitto__strdup((char *) ASN1_STRING_data(name_asn1));
+					context->username = mosquitto__strdup((char *) ASN1_STRING_get0_data(name_asn1));
 					if(!context->username){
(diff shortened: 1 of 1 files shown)


yoonghm:
I have signed the ECA

ralight:
Thank you for looking at this. As I recall, the reason this hasn't been changed so far is compatibility with earlier versions of openssl. Do you know what version of openssl introduced `ASN1_STRING_get0_data()`?

ralight:
It may be that we want a version specific test here, depending on when the new function became available - if so, then you can update your PR like this:

```
# make changes
git add src/handle_connect.c
git commit --amend
git push -f
```

Note that bug fixes (which this is, strictly speaking) should be based on the `fixes` branch, but it's not a big deal here because I can rebase it on it.

yoonghm:
Hi,

  It is version 1.1.1.  I did it via github, hence I could not perform the
command as required by eclipse. Thanks.

PR #1018, opened 2018-11-06
DYS-173: Adding support for QNX7.0.0

w0land:
This change allows mosquitto to be compiled and used on QNX platform (qnx700). 
My custom CMake toolchain adds CMake variable ``QNX`` ; additionally qcc (QNX compiler) adds ``__QNX__``  macro that can be used in sources. 

Signed-off-by: Bartosz Taczała <bartosz.taczala@mobica.com>

-----


client/CMakeLists.txt
```diff
@@ -11,8 +11,14 @@ endif (${WITH_SRV} STREQUAL ON)
...
-target_link_libraries(mosquitto_pub libmosquitto)
-target_link_libraries(mosquitto_sub libmosquitto)
+set(extra_libraries "")
+
+if (QNX)
+    set (extra_libraries socket)
+endif()
+
+target_link_libraries(mosquitto_pub libmosquitto ${extra_libraries})
+target_link_libraries(mosquitto_sub libmosquitto ${extra_libraries})
```
karlp: according to cmake documentation, you can just call target_link_libraries multiple times.  Just leaving the existing lines, and doing 
```
if (QNX)
   target_link_libraries(mosquitto_pub socket)
   target_link_libraries(mosquitto_sub socket)
endif()

File: src/CMakeLists.txt [hunk 2]
```diff
@@ -115,7 +115,10 @@ endif (WIN32 OR CYGWIN)
 
 add_definitions (-DWITH_BROKER)
 
-set (MOSQ_LIBS ${MOSQ_LIBS} ${OPENSSL_LIBRARIES})
+set (MOSQ_LIBS ${MOSQ_LIBS})
+if(WITH_TLS)
```
karlp: this seems like a standalone bug fix? ie, valid, but out of scope for a QNX support commit

src/CMakeLists.txt
```diff
@@ -126,17 +129,20 @@ if (HAVE_GETADDRINFO_A)
...
 	if (APPLE)
 		set (MOSQ_LIBS ${MOSQ_LIBS} dl m)
-	else (APPLE)
-		set (MOSQ_LIBS ${MOSQ_LIBS} dl m)
-		find_library(LIBRT rt)
-		if (LIBRT)
-			set (MOSQ_LIBS ${MOSQ_LIBS} rt)
-		endif (LIBRT)
-	endif (APPLE)
+    else()
```
karlp: why not just an elseif instead of nesting further here? would leave the positive logic in place too, if apple, if qnx, else, rather than mixing positive and negative logic.

src/CMakeLists.txt
w0land: Ok, I'll create another PR for this

karlp:
looks sane to me, (one whitespace new line added) but I've not actually tested this or targetted qnx.

More interestingly, do the makefiles work? the cmakefiles are, per readme, only intended for windows /mac "On Windows and Mac, use cmake to build. On other platforms, just run make to build."

ralight:
I'm happy to take this as is, thanks for the contribution. Way back in the day (around 2012) I compiled mosquitto for QNX, but haven't touched it since. It's nice to see some more interest from that side.

Like Karl says, the bare Makefiles are intended as the canonical build scripts for the project, Mac and Windows not withstanding. I don't suppose it matters hugely, but it would be nice to hear if the build works with them as well.

w0land:
Oh.. Somehow I didn't catch that. 
Makefiles doesn't work as they are trying to link againt rt and pthread that does not exists under QNX (functionality is simply a part of QNX's libc)

ralight:
There's nothing inherently wrong with using the cmake files, but the others are considered the best supported. For QNX I wouldn't worry particularly.

Issue #1075, opened 2018-12-12
Specifying secure and non-secure listeners causes immediate exit with code 1


sirockin:
Docker image eclipse-mosquitto:latest (1.5).
Host OS: Linux instance-1 4.14.67+

mosquitto.conf excerpt:
```
# Uncommenting the next line makes server exit with error code 1
# listener 1883 # Unencrypted

# mqtt/tls
listener 8883
cafile /certificate/ca_bundle.crt
certfile /certificate/certificate.crt
keyfile /certificate/private.key

# listener 1884   # Non-encrypted

listener 9001
protocol websockets
cafile /certificate/ca_bundle.crt
certfile /certificate/certificate.crt
keyfile /certificate/private.key

```

Output with 'listener 1883' line uncommented:
```
mqtt-broker_1  | 1544611789: mosquitto version 1.5.5 starting
mqtt-broker_1  | 1544611789: Config loaded from /mosquitto/config/mosquitto.conf.
gclouddeploy_mqtt-broker_1 exited with code 1
```

ralight:
My guess is that the # is causing problems. # is valid and expected input for some of the options so it is only valid as a comment when on the very first character of the line.

If that is the problem, I'll leave this open to see if there can be anything done to improve it - especially the error reporting.

sirockin:
Yes that worked - thanks!

kumar003vinod:
Hi @ralight 
I have proposed one solution for this with #1091. Please take a look.

ralight:
Closing this thanks to the PR from @kumar003vinod !

PR #1091, opened 2018-12-27
ignore inline comments while parsing optional config params

kumar003vinod:
This is a fix I am proposing for #1075 . The inline comment does not work for configuration parameters which have optional values. like `listener port [bind address/host]`.
In such cases currently we read inline comment as optional parameter value. This patch just ignores the inline comments and default value of optional params is used.
In this patch I have considered `listener`, `address` and `topic` as these have optional parameters in configuration file (I hope I did not miss any).

Other approach could be to display proper error message, but for other configuration keys where we do not have optional parameters we end up ignoring inline comments, because we do not parse the line once we get the required values, so this will make it consistent across the whole config.

This will not work if we have optional parameter which can start from `#`, but in this case we do not have such optional parameter.

Signed-off-by: Vinod Kumar <kumar003vinod@gmail.com>

-----

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -765,6 +765,9 @@ int config__read_file_core(struct mosquitto__config *config, bool reload, struct
 						return MOSQ_ERR_INVAL;
 					}
 					while((token = strtok_r(NULL, " ", &saveptr))){
+						if (token[0] == '#'){
+							break;
+						}
 						cur_bridge->address_count++;
 						cur_bridge->addresses = mosquitto__realloc(cur_bridge->addresses, sizeof(struct bridge_address)*cur_bridge->address_count);
 						if(!cur_bridge->addresses){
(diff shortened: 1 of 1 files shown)


ralight:
Thanks for taking a look at this.I'm trying to think if there is a problem where only the options with optional parameters have the option to have a comment. As it stands, the only place a comment is valid is with # at the start of a line. With your change (which is an improvement), some options now have the ability to have a comment after them, but not all.

In some cases this is not going to cause confusion. If you type `persistence true # enable saving to disk`, then you will get an error about an invalid value. This isn't *friendly* though.

Would you consider expanding your work to adapt the `conf__parse_ssize_t()`, `conf__parse_int()`, and `conf__parse_bool()` functions to account for comments? That would then cover a lot more of the possible options.

kumar003vinod:
Hi @ralight , 
`persistence true # enable saving to disk` works fine, the invalid value error is thrown when there is no space before `#`, like `persistence true# enable saving to disk`.
In case of configuration parameters with optional values, we were reading comment as value, which will be solved by this. In these configuration parameters also if space is not given before `#`, invalid value error is thrown.

ralight:
Oh yes, you're right. It actually treats *anything* after the `persistence true` as a comment, whether with a `#` or not. This could be regarded as a bug as well of course!

kumar003vinod:
Should we extend scope of this PR to include this handling (to raise error in case we have any non comment after we have parsed required config) ?, or should we consider this as a new improvement and I can provide a separate PR for this.

ralight:
Let's keep things simple and separate. This work stands on its own, if you'd like to make a new PR as you suggest please go ahead :)

Issue #339, opened 2016-12-22
Retrieve websocket client address from X-Real-IP/X-Forwarded-For headers


brandond:
LWS has had support for X-Real-IP and recently added X-Fowarded-For: https://github.com/warmcat/libwebsockets/commit/64ee08a895208d65e5869a9f413da67c09fa3ac8

Mosquitto should have an option to retrieve the contents of one of these headers, instead of calling `getpeername` by way of `_mosquitto_socket_get_address`. This needs to be done during the HTTP part of the handshake, as documented here:
https://libwebsockets.org/lws-api-doc-master/html/group__HTTP-headers-read.html

ralight:
That sounds reasonable to me.

robymus:
I have a preliminary implementation which I use now in our systems:

https://github.com/robymus/mosquitto/commit/c5cfe68ae722ceb98e4f2ad64875e04f5582f27a

However, a configuration option would be nice, to use these headers only when requested or when the source ip is matching a pattern (localhost, or local network, or proxy address?) - it is alright and useful, when the websocket connection is passed through a reverse proxy, but when it is directly on the internet, it opens up the system for ip address spoofing.

Could you give some pointers what configuration options would be reasonable, to be in sync with existing ones?

DConcord:
This would be really nice. @ralight is this doable?

ralight:
This is now suppoerted in the `develop` branch for builtin (non-libwebsockets) websockets.

PR #1111, opened 2019-01-20
Retrieve websocket client address from X-Real-IP/X-Forwarded-For headers

Auswaschbar:
Using the patch that robymus provided as a base in Issue #339, I went ahead and added a config variable as was suggested in the issue.

It works pretty well, but at least on my system with libwebsockets-3.0.1 there is a small formatting issue: lws_hdr_copy() includes the colon and the space of the header, so the IP looks like ": 127.0.0.1" in the log. I am unsure if this is a bug in lws or intentional, so I didn'T know how to proceed and just left it as it was.

### DIFF
--- a/man/mosquitto.conf.5.xml
+++ b/man/mosquitto.conf.5.xml
@@ -846,6 +846,17 @@
 							Defaults to 0.</para>
 					</listitem>
 				</varlistentry>
+				<varlistentry>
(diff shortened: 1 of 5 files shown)


NorthernMan54:
Not to dig out a 3 year old pull request, but was this ever implemented ?  Looking at the current master and I don't see it in the code.

hjhjdev:
Seems like it didn't implemented yet

PR #1132, opened 2019-01-30
Fix missing AF_INET definition

stlawrance:
Without this, mosquitto doesn't build on FreeBSD with websockets enabled

Signed-off-by: Steven Lawrance <stl@koffein.net>

### DIFF
--- a/src/websockets.c
+++ b/src/websockets.c
@@ -41,6 +41,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 #include <stdlib.h>
 #include <errno.h>
+#include <sys/socket.h>
 #include <sys/stat.h>
 
 extern struct mosquitto_db int_db;


karlp:
This should have been fixed by https://github.com/eclipse/mosquitto/commit/0368a8c01d2726c81624b1b08a096c307dbb45ab in the fixes branch?

stlawrance:
> This should have been fixed by [0368a8c](https://github.com/eclipse/mosquitto/commit/0368a8c01d2726c81624b1b08a096c307dbb45ab) in the fixes branch?

That's a different source file so no, I guess it's a different problem.  Curious that I didn't hit it though.  `socket.h` should be included for AF_INET anyway, but I haven't looked at how it winds up defined on other builds.

ralight:
Thanks, this looks a good catch.

PR #1136, opened 2019-01-31
Allow binding a listener to a specific network interface

stlawrance:
This is for restricting connections to a specific interface, where that interface may have multiple or changing addresses.

Signed-off-by: Steven Lawrance <stl@koffein.net>

### DIFF
--- a/man/mosquitto.conf.5.xml
+++ b/man/mosquitto.conf.5.xml
@@ -729,6 +729,16 @@
 					</listitem>
 				</varlistentry>
 				<varlistentry>
+					<term><option>bind_interface</option> <replaceable>device</replaceable></term>
+					<listitem>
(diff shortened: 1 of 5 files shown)


stlawrance:
It should work on all Linux systems, where it's not a new option (I think since 2.0 or 2.2).  I've tested that it builds on FreeBSD and produces an error if you attempt to configure it, which I guess is better than just a warning since it's a security-related option.  Unfortunately there doesn't appear to be an equivalent option on FreeBSD or Windows, so they will have to make do with the bind_address.

I've added the option to the sample `mosquitto.conf` including a note that it only works on Linux.

ralight:
Thanks for the document updates, I've now merged this via 208c3d3e85788364402d29b9f7c5963fba7cc375.

PR #1142, opened 2019-02-03
FIX #1080

### LINKED ISSUE #1080: Broker can't restart with log file on Windows
Hello,
Running on Windows (in service mode) with logging configuration to file  : "log_dest file ..." in configuration file, when stop the service, the service can't restart.
By reading the code,  in util_mosq.c l.422, "CreateFile" function is called with "CREATE_NEW" parameter.
In cas of an existinf log fille, INVALID_HANDLE_VALUE is returned , this case is not tested in code.
I suggest to use "CREATE_ALWAYS" parameter instead.

Not tested, but problem is probably also present in "application" mode.

lcouz:
Fix proposal for : Broker can't restart with log file on Windows #1080

### DIFF
--- a/lib/util_mosq.c
+++ b/lib/util_mosq.c
@@ -421,7 +421,7 @@ FILE *mosquitto__fopen(const char *path, const char *mode, bool restrict_read)
 
 			hfile = CreateFile(buf, GENERIC_READ | GENERIC_WRITE, 0,
 				&sec,
-				CREATE_NEW,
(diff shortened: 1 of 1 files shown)


ralight:
Thanks for this PR, but I'm not sure it quite goes far enough because it doesn't distinguish between creating a new file, appending to an existing one, or opening an existing one. I've made a change that covers those cases as well so I'm going to close this. Thanks very much for the hint.

PR #1151, opened 2019-02-09
add check for inline comments in config parsing

kumar003vinod:
As suggested with #1091, I propose this new PR.
Current config parsing treats anything after required config parameters as a comment. In this change, I parse the config line further to make sure that we only have comment text after the required configuration.
With this change assumed that `log_dest` `include_dir`, `bridge_protocol_version` and `auth_opt_*` config parameters do not have space in configuration value.

For example, according to old logic `max_connections 100 comment` configuration is valid, but with this change, we will throw an error and `max_connections 100 # comment` is valid.
```
Error: Invalid configuration (comment) found after 'max_connections' configuration value (100).
```

And added one minor fix where current parsing logic throws an error if some a comment line starts with space.
```
# this is a comment
 # this is another comment
```

Signed-off-by: Vinod Kumar <kumar003vinod@gmail.com>

-----

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -71,6 +71,7 @@ static int conf__parse_string(char **token, const char *name, char **value, char
 static int config__read_file(struct mosquitto__config *config, bool reload, const char *file, struct config_recurse *config_tmp, int level, int *lineno);
 static int config__check(struct mosquitto__config *config);
 static void config__cleanup_plugins(struct mosquitto__config *config);
+static int config__check_for_inline_comment(char **token, const char *name, char* saveptr);
 
 static char *fgets_extending(char **buf, int *buflen, FILE *stream)
 {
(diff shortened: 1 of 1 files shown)


ralight:
I've been considering this for a while and I think that it is important not to break existing configs. It is possible that  `log_dest` `include_dir`, and `auth_opt_*` all have spaces in legitimate cases at the moment. That only leaves `tls_version`. I think it might be best to leave this until breaking changes in version 2.0.

kumar003vinod:
Thanks for your input. If I remove the check for these config parameters, will it be good to go?
In this PR I also did not handle for `conf__parse_string`, because it also allows spaces in values, maybe we can come up with a better solution for these cases.

ralight:
I think I'd prefer to leave it as is for now actually. There's too much aggravation for people if something breaks. Breaking changes in 2.0 could include a change in config file format to something more standard, which removes this issue.

kumar003vinod:
Thanks, Roger. That makes perfect sense.

PR #1182, opened 2019-02-27
Fix comparison of boolean values in CMake build.

tgurr:
Follow-up of 70c4097b6ff165e40ff5538700e90ec8e2184ce8.

-----


ralight:
Thank you... not quite sure how I didn't notice they were missing before.

PR #1216, opened 2019-03-30
Let auth plugin handle if no username matched

### LINKED ISSUE #1215: Auth plugin ALWAYS not receiving callback when using password_file
As per https://mosquitto.org/man/mosquitto-conf-5.html
"If password_file, or acl_file are used in the config file alongsize auth_plugin, the plugin checks will run after the build in checks."

https://github.com/eclipse/mosquitto/blob/2d360fd49521ef8c1694288023f93cc605697189/src/security_default.c#L833-L886

In above code, if no user is found (having a hash table), `mosquitto_unpwd_check_default()` return `MOSQ_ERR_AUTH` (the last line of function)

shouldn't it actually return `MOSQ_ERR_PLUGIN_DEFER` (at last line) since it suppose to defer it to other as per the docs)?

`mosquitto_auth_unpwd_check()` will never receive the callback if `password_file` is used in `mosquitto.conf`

Note: im very new to the code and based on heuristic im trying to guess the reason my plugin is not receiving callback.

```
diff --git a/src/security_default.c b/src/security_default.c
index 99b7809..8f896b0 100644
--- a/src/security_default.c
+++ b/src/security_default.c
@@ -882,7 +882,7 @@ int mosquitto_unpwd_check_default(struct mosquitto_db *db, struct mosquitto *con
 		}
 	}
 
-	return MOSQ_ERR_AUTH;
+	return MOSQ_ERR_PLUGIN_DEFER;
 }
 
 static int unpwd__cleanup(struct mosquitto__unpwd **root, bool reload)
```

kuldeepdhaka:
mosquitto_unpwd_check_default() should return MOSQ_ERR_PLUGIN_DEFER instead MOSQ_ERR_AUTH
if no username matched so that auth plugin can handle it.

According to docs:
"If password_file, or acl_file are used in the config file alongsize auth_plugin,
the plugin checks will run after the build in checks."

Fixes #1215

-----

### DIFF
--- a/src/security_default.c
+++ b/src/security_default.c
@@ -882,7 +882,7 @@ int mosquitto_unpwd_check_default(struct mosquitto_db *db, struct mosquitto *con
 		}
 	}
 
-	return MOSQ_ERR_AUTH;
+	return MOSQ_ERR_PLUGIN_DEFER;
 }
 
 static int unpwd__cleanup(struct mosquitto__unpwd **root, bool reload)


ralight:
This possibly needs changing, thanks for noting it. There is no way I can accept any changes to this without extensive tests though.

PR #1234, opened 2019-04-19
Fix comment for option values

krestivo-kdinfotech:
Fix comment, these are option values, not error values.


ralight:
Thank you very much!

PR #1254, opened 2019-05-01
Fix password input when MAX_BUFFER_LEN is not 1024 in mosquitto_passwd.c

ccvca:
This handles some inconsistencies in mosquitto_passwd.c . The MAX_BUFFER_LEN was not used at all places, which could result in a buffer overflow in case of MAX_BUFFER_LEN != 1024 here:
https://github.com/eclipse/mosquitto/blob/995f90dbafa82c7fd3d24f7174d094f11c2d6c85/src/mosquitto_passwd.c#L529

Also `strncpy` does not garantee to produce a Null-terminated string in destination, if src is longer than the third parameter `len`:
https://github.com/eclipse/mosquitto/blob/995f90dbafa82c7fd3d24f7174d094f11c2d6c85/src/mosquitto_passwd.c#L367

Due to the special case that MAX_BUFFER_LEN is set to 1024, this is not a critical bug, but now it should be safe to change MAX_BUFFER_LEN if necessary.

As I do not have a working OpenSSL-Setup, I'm unable to test it myself. But I hope the CI will do it.

-----

### DIFF
--- a/src/mosquitto_passwd.c
+++ b/src/mosquitto_passwd.c
@@ -339,13 +339,15 @@ int gets_quiet(char *s, int len)
 #endif
 }
 
-int get_password(char *password, int len)
+int get_password(char *password, size_t len)
 {
 	char pw1[MAX_BUFFER_LEN], pw2[MAX_BUFFER_LEN];
+	size_t minLen;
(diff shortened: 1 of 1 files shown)


ralight:
Thank you, I agree this isn't super critical but definitely good to fix. Sorry it's taken so long to merge.

PR #1276, opened 2019-05-13
Develop

jsaak:
added upgrading instructions to man mosquitto


ralight:
This is a great suggestion, but I've written it slightly differently and put it alongside the `persistence` option in the config man page. Does that look ok?

jsaak:
Looks good, closing this now.

PR #1308, opened 2019-06-06
No pthread_cancel() on Android

dandrader:

-----

pthread_cancel() is not available on Android

Thus mosquitto_loop_start() and mosquitto_loop_stop()  won't be available there (and mosquitto_connect_async() as a consequence).

### DIFF
--- a/config.h
+++ b/config.h
@@ -60,4 +60,9 @@
 
 #define UNUSED(A) (void)(A)
 
+/* Android Bionic libpthread implementation doesn't have pthread_cancel */
+#ifndef ANDROID
+#  define HAVE_PTHREAD_CANCEL
+#endif
+
 #endif
--- a/lib/mosquitto.c
+++ b/lib/mosquitto.c
@@ -201,11 +201,13 @@ void mosquitto__destroy(struct mosquitto *mosq)
 	if(!mosq) return;
 
 #ifdef WITH_THREADING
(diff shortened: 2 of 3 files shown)


ralight:
Thanks for this Dan, I wasn't aware of that problem on Android. I hope you don't mind, I've fixed this in a slightly simpler fashion in the commit I've just pushed to `fixes`.

dandrader:
Hi Robert,

Please correct me if I'm wrong, but If you unset WITH_THREADING then the mutexes won't be operational and therefore the API will no longer be thread-safe, right?

In my usage I'm running mosquitto_loop() in one thread, and the rest of the API in another one. Without calling mosquitto_threaded_set(... true) I get the obscure crashes mentioned in its documentation.

Disabling just the pthread_cancel usage, like I did in this patch, you lose the mosquitto_connect_async API but still keep using real mutexes to guard the internal data structures.

Will mosquitto_threaded_set(... true) still keep its promises without WITH_THREADING defined?

ralight:
You make a good point, let's go with your solution instead.

nihilus:
Actually bionic libc lacks pthread_cancel() as you've figured out; it requires stack unwinding and is to be considered a "hack" in most libc-implementations (no safe cancellation points), however for mosquitto I personally find waiting for pthread_join() more feasible than disable threading. Code based on the async-API won't be able to run on Android which is a pity.

PR #1328, opened 2019-07-02
Install pkg-config files into arch dependent locations also for CMake builds

tgurr:
libmosquitto.pc and libmosquittopp.pc contain arch dependent information.
The Makefile based build already does this correctly.

Signed-off-by: Timo Gurr <timo.gurr@gmail.com>


ralight:
Thank you! Sorry it has taken so long to get round to merging this.

PR #1331, opened 2019-07-04
Support tilde character as HOME in config paths

sealor:

-----

This feature provides path expansion for the log file path when reading the mosquitto broker configuration. It expands the tilde character (~) to the home directory of the current user.

e.g.
```
log_dest file ~/log/mosquitto.log
```

### DIFF
--- a/man/mosquitto.conf.5.xml
+++ b/man/mosquitto.conf.5.xml
@@ -36,6 +36,9 @@
 		comment.</para>
 		<para>Configuration lines start with a variable name. The variable
 		value is separated from the name by a single space.</para>
+		<para>All file and directory paths support path expansion. If the path
+		starts with <code>~/</code> it is substituted with the value of the
+		environment variable <code>HOME</code>.</para>
 	</refsect1>
 
 	<refsect1>
--- a/src/conf.c
+++ b/src/conf.c
@@ -67,6 +67,7 @@ static int conf__parse_bool(char **token, const char *name, bool *value, char *s
 static int conf__parse_int(char **token, const char *name, int *value, char *saveptr);
 static int conf__parse_ssize_t(char **token, const char *name, ssize_t *value, char *saveptr);
 static int conf__parse_string(char **token, const char *name, char **value, char *saveptr);
+static int conf__parse_path(char **token, const char *name, char **value, char *saveptr);
 static int config__read_file(struct mosquitto__config *config, bool reload, const char *file, struct config_recurse *config_tmp, int level, int *lineno);
 static int config__check(struct mosquitto__config *config);
 static void config__cleanup_plugins(struct mosquitto__config *config);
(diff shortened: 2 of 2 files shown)


man/mosquitto.conf.5.xml
```diff
@@ -427,7 +427,8 @@ D.conf
 						e.g. "log_dest file /var/log/mosquitto.log". The file
 						will be closed and reopened when the broker receives a
 						HUP signal. Only a single file destination may be
-						configured.</para>
+						configured. A leading tilde character is substituted by
+						the content of environment variable HOME.</para>
```
MichaelCook:
```suggestion
						the value of the environment variable HOME.</para>
```

src/conf.c
```diff
@@ -1465,7 +1465,19 @@ int config__read_file_core(struct mosquitto__config *config, bool reload, struct
 								token++;
 							}
 							if(token[0]){
-								config->log_file = mosquitto__strdup(token);
+								const char* home = getenv("HOME");
```
MichaelCook: There's a path through this code where `config->log_file` doesn't get assigned.
```suggestion
								config->log_file = NULL;
```

sealor: @MichaelCook
Thank you for your very fast response!
I added two commits to apply your suggestions. I also could squash them into one commit if needed.

ralight:
I fear the implementation of this may be a bit trickier than what you've got. At the moment you only look for `~` in the first character. If I use `~roger/mosquitto.log` then I'll get unexpected results. It's not unreasonable to assume that if `~/` works then `~roger/` would work as well. Going down the full route of all user paths being expanded is a lot of complexity I wouldn't want to get into just for tilde substitution.

As it stands, I'm not sure about whether this is something to merge. As a minimum it would have to check for `~/`, not just `~`. I would also want it to work for other config paths as well. I need a bit of convincing that this is widely worthwhile as well, so go on, persuade me.

For reference, if you do pull requests in the future here or anywhere else, and you have to modify them, then you can do the following:

```
# Make changes needed
git add file.c
git commit --amend
git push -f
```

This updates the existing commit and the pull request copes with it correctly.

sealor:
@ralight Thank you for your feedback.

You are absolutely right. We implemented all your suggestions in the updated PR.

In our current setup we are not able to use a relative path in the mosquitto configuration
because we can not control from which current directory it will be started.
We also cannot use an absolute path because the broker with the same configuration is started
on different machines with different users.

We already implemented a workaround with a shell script as wrapper which changes the current
directory to the users HOME and use a relative path in the configuration.

But we think it would be nice if mosquitto could provide path expansion. We are sure this
feature would also be interesting for others too.

Assar63:
Is this commit also considering the "user" in mosquitto.conf?

ralight:
On balance I don't think this is a good approach because it's not cross platform. I think the better approach would be to allow environment variables.

PR #1392, opened 2019-08-26
Add missing EXPOSE instruction to all Dockerfiles

heikobornholdt:


ralight:
I'm not sure this is the correct thing to do. 1883 and 9001 are given as examples in the readme, but aren't actually configured by default. You can always expose the ports that you are using `-p`, as described in the readme. If there is a reason why this EXPOSE is needed beyond what I've said, please do explain.

heikobornholdt:
It is common for Docker Images to expose the default ports in this way (see [nginx](https://github.com/nginxinc/docker-nginx/blob/e3bbc1131a683dabf868268e62b9d3fbd250191b/mainline/buster/Dockerfile#L99), [redis](https://github.com/docker-library/redis/blob/0b2910f292fa6ac32318cb2acc84355b11aa8a7a/5.0/Dockerfile#L106), or [mysql](https://github.com/docker-library/mysql/blob/4af273a07854d7e4b68c5148b8e23b86aa8706e2/8.0/Dockerfile#L74)). Mosquitto uses these ports when no configuration is specified.

I would like to use the Mosquitto Docker image as a [GitLab CI service](https://docs.gitlab.com/ee/ci/services/). GitLab can find the ports from the image with zero configuration and expose them automatically. This feature requires the `EXPOSE` instruction in the Dockerfile. Since GitLab does not support the configuration of the services, it is currently not possible for me to use Mosquitto in GitLab CI.

heikobornholdt:
Other programs also depend on the `EXPOSE` instruction. For example the popular program [Kitematic](https://kitematic.com), which offers a GUI to start docker containers.

ralight:
Thanks for the explanation. The images only use port 1883 by default, so I think the EXPOSE instructions should only include that port. If you update the PR to be just port 1883 then I will merge it, otherwise I am happy to do it myself and attribute you.

heikobornholdt:
Wonderful, I have updated the PR

ralight:
Thank you!

PR #1405, opened 2019-09-06
Fix usage of tls_engine and tls_keyform.

woelfman:
The current implementation does not properly forward the engine
parameters to OpenSSL causing OpenSSL to incorrectly attempt to open the
engine key uri as a file.

This fix enables net_mosq.c:net__init_ssl_ctx to use ENGINE_load_private_key instead of SSL_CTX_use_PrivateKey_file.

Signed-off-by: Matt Woelfel <matt@woelfware.com>


ralight:
Is that the whole commit? It doesn't quite match up to the comment for me.

woelfman:
Yes, this is it.  Some of the engine parameters specified in mosquitto.conf are not making it all the way to where they are being used.

ralight:
Ok, thanks I'll look at it properly soon.

ralight:
Thanks very much!

PR #1473, opened 2019-10-24
mosquitto_pub: split main loop.

majekw:
Main loop has two different logic inside: one for stdin input,
second for all other cases. Exit loop condition is also different
for both variants. This commit splits these two variants into
two separate loops.
It could help fixing other issues in this part of code and greatly improve it's readability.
There are also a little clean-up things:
- use `rc` in stdin loop, so we get exit status and spare few bytes
- use MOSQ_ERR_SUCCESS instead of cryptic `0`

Signed-off-by: Marek Wodzinski <majek@w7i.pl>

-----

### DIFF
--- a/client/pub_client.c
+++ b/client/pub_client.c
@@ -223,7 +223,7 @@ int pub_shared_loop(struct mosquitto *mosq)
 {
 	int read_len;
 	int pos;
-	int rc, rc2;
+	int rc;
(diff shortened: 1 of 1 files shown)


ralight:
I like this, and I'm actually going to take it a step further to split those loops into separate functions to make it even more readable. Thanks very much!

PR #1508, opened 2019-11-22
allow client to publish on many topics

velimir:
Hi,

Would you be interested in accepting the following feature? If so I would work on getting the tests done.

### DIFF
--- a/client/client_shared.c
+++ b/client/client_shared.c
@@ -410,7 +410,7 @@ int cfg_add_topic(struct mosq_config *cfg, int type, char *topic, const char *ar
 		fprintf(stderr, "Error: Malformed UTF-8 in %s argument.\n\n", arg);
 		return 1;
 	}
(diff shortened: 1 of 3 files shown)


karlp:
Perhaps describe what the goal is here, and the behaviours that can't be managed with existing tooling?

velimir:
The goal here is to allow `mosquitto_pub` client to publish on multiple topics:

```
mosquitto_pub -m "hello world" -t client/updates/0 -t client/updates/1
```

The message is sent on the topics. Client cycles through the topics sending the provided message to the given topics one by one.

This is very handy for profiling a broker, where a client can publish on many topics within the same connection.

It can be done with existing tooling as there's no option at the moment to use multiple topics for the same connection.

ralight:
I don't believe this is the direction that mosquitto_pub should go in. It would be potentially confusing to its core purpose.

PR #1522, opened 2019-12-01
Support for unix domain socket based connection passing

qubeck:
This feature adds support for unix domain socket based connection passing to allow an other process to pass already accepted connections to a running instance of mosquitto broker.

In case a unix domain socket listener is being used this patch evaluates additionally the first incoming message for presence of ancillary data of type SCM_RIGHTS to take a passed connection file descriptor and use it instead of the original accepted socket.

(This pull request makes #771 obsolete)

-----
-----

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -118,6 +118,30 @@ int net__socket_accept(struct mosquitto_db *db, mosq_sock_t listensock)
 	struct request_info wrap_req;
 	char address[1024];
 #endif
+#if defined WITH_UNIX_SOCKETS && defined WITH_UDS_CP
+	struct fwd_con_ctx
+	{
+		uint8_t reserved;
+	} uds_fwd_context;
+	char control[CMSG_SPACE(sizeof(int))];
+	struct iovec  iov = { .iov_base = &uds_fwd_context, .iov_len = sizeof(uds_fwd_context) };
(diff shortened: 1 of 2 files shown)


karlp:
You're going to have to (at least) update the makefile too, as they're canonical for non-windows systems, not cmake

src/net.c
```diff
@@ -118,6 +118,30 @@ int net__socket_accept(struct mosquitto_db *db, mosq_sock_t listensock)
 	struct request_info wrap_req;
 	char address[1024];
 #endif
+#if defined WITH_UNIX_SOCKETS && defined WITH_UDS_CP
+	struct fwd_con_ctx
+	{
+		uint8_t reserved;
```
karlp: what's this for?

qubeck: since at least one byte needs to be sent along the way I just wanted to reflect that in the receiving iov buffer. That's all. At least for the receiving site iov can be NULL and length can be changed to zero ofcourse... Just wanted to keep consistency and annotate the possibility to forward other context related information.

Should I change that and recreate the PR with necessary Makefile update?

qubeck:
> You're going to have to (at least) update the makefile too, as they're canonical for non-windows systems, not cmake

Thx, I'll provide integration to the non-cmake based makefile build

Issue #1509, opened 2019-11-23
Using native threads on Windows


laudrup:
Currently mosquitto requires pthread for Windows to use threading on Windows.

Is there any reason why this library doesn't use some simple wrappers for the threading functionality it requires so it can use native threads on both Posix and Windows?

It seems like it should be relatively simple to implement and I will be happy to do so through a pull request, just want to know if there's any reason why this hasn't been done before I spent time on it.

Sorry if this is more of a question than an actual issue. Don't know if there's a better way to ask questions like this on github.

ralight:
Libmosquitto doesn't absolutely rely on pthreads for Windows, it just needs an implementation of the pthreads API, so you are quite correct that you could write a simple wrapper.

There is a BSD licensed implementation that attpts what you suggest here: http://locklessinc.com/articles/pthreads_on_windows/

laudrup:
Thanks for your reply.

So would you be interested in a pull request for a wrapper that uses native threads on Windows?

crazydef:
Hi,

I've just started looking into integrating mosquitto into our codebase, and this is a major blocker for us (who writes single threaded applications these days?).

What's the status of this?  (I see your pull request is failing on Mac for seemingly unrelated reasons - is anyone looking into that?  Can that be worked around?)

laudrup:
Hi @crazydef,

The master branch and develop branch is indeed currently failing so that is not related to my pull request at all.

Not sure what the status of this pull request is. I think the first thing that should be fixed is the master and develop branch. Before that no pull requests can really be merged (that would defeat the whole point of having CI checks IMO).

There doesn't seem to be much development on this project at all, not sure what to make of that...

crazydef:
Well, two months since the last push is definitely better than some github projects I've seen recently.  :)

But either way, I ended up using the single-threaded version of mosquitto and marshalled all incoming and outgoing messages through my own message queues and have a worker thread that calls into mosquitto manually to pump its message loop and get things moving.

Ironically, it turned out there are other parts of our system that also needed things to happen on specific threads, so doing it this way actually turned out a bit better for me in the end.  Barring any major bugs or issues though, I don't think I'll be looking at this part of the system again anytime soon.  :)

karlp:
re activity, perhaps look at https://github.com/eclipse/mosquitto/commits/fixes instead. even two months would just mean "stable" not "this project is dead"

laudrup:
[...]

PR #1528, opened 2019-12-10
Windows thread wrappers

laudrup:
This is an attempt to address what I mentioned in #1509.

I would like to go further with this if it has any interested.

It has been a bit hard to make any real test of this, since currently the tests don't run natively on Windows. I would like to help improving that as well if it has any interest.

Please let me know what you think. I definitely do not expect you to merge this as it is.


laudrup:
Sorry for force pushing :-(

raspopov:
IMHO I don't think it's useful since **pthreads** already is a wrapper of Windows threads: git://git.code.sf.net/p/pthreads4w/code
I tested it and it's works well although has memory leaks by design (due thread handlers reusing).

ralight:
Thank you for your efforts on this. I decided to implement things differently by creating the `pthread_*` functions using the appropriate Windows functions. I think it is simpler this way. It will be part of the 2.1 release. Thanks again for your input.

PR #1542, opened 2019-12-18
fix: replace sleep with (p)select in loop_forever

ChristianS99:
sleep was blocking loop_stop(force=false) since it
was uniteruptible

Signed-off-by: Christian Schneider <cschneider@radiodata.biz>

 --> How?

-----


ralight:
Thank you! And sorry it's taken so long to get to.

karlp:
nanosleep is interruptible, was this actually only a problem on windows?  (Just asking as I'm using nanosleep in lots of places to avoid having to deal with select)

ChristianS99:
maybe uninteruptable isn't quite on the point. What I meant was, that it didn't get canceled, when loop_stop(force=false) was called, actually i didn't look at nanosleep, if it can get interupted, but used the select approach.

PR #1568, opened 2020-01-22
Bugfix: include "deps" directory only if BUNDLED_DEPS has been provid…

LocutusOfBorg:
.


ralight:
Thanks, that's a good catch.

PR #1569, opened 2020-01-22
Bugfix: enabling DLT was overriding everything else on linker flags b…

LocutusOfBorg:
…ecause of error in cmake set keyword


ralight:
Thank you.

PR #1571, opened 2020-01-22
Add dynamic symbols linking with cmake too

LocutusOfBorg:
.


LocutusOfBorg:
hello @ralight ping please?

rzr:
Ping ?

For the record 3fe5468f1bdca1bff1d18cf43c9e338f41aa9e32 was applied in debian packaging

rzr:
For the record this patch has also been applied to OE:

https://github.com/openembedded/meta-openembedded/blob/master/meta-networking/recipes-connectivity/mosquitto/files/1571.patch

PR #1572, opened 2020-01-22
Move travis to cmake builds for linux too

LocutusOfBorg:


ralight:
Sorry, I'm not interested in this change. We've discussed before how the Makefiles are the canonical build and test system.

PR #1573, opened 2020-01-24
Memory leak in handle_unsubscribe.c

panava:
Reason: In line 70, the memory allocation for the pointer reasons_codes may
result to a memory leak due to the many returns (e.g as the one in line 78)
occuring in the program's path until reaching the mosquitto__free at line 122.

Fix: I moved the memory allocation code block (lines 69-73) just before
the line 102. This is the first place the pointer reason_codes is used, while
the following mosquitto__free operators free the allocated memory correctly.

Signed-off-by: Panagiotis Vasilikos <panagiotis.vasilikos@alexandra.dk>

-----

### DIFF
--- a/src/handle_unsubscribe.c
+++ b/src/handle_unsubscribe.c
@@ -66,12 +66,6 @@ int handle__unsubscribe(struct mosquitto_db *db, struct mosquitto *context)
 		}
 	}
 
-	reason_code_max = 10;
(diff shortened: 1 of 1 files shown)


karlp:
Is this meant to be closed or open?

panava:
I had an error in my fix for this particular pull, so I closed it and I made a new pull request with the correct changes in the code.

PR #1577, opened 2020-01-29
Potential Null pointer dereference in util_mosq.c

panava:
Reason: There is no check that the mosquitto__malloc at line 162 was sucessfull.
This could result to a Null pointer dereference in the memcpy call at line 166.

Fix: I added the check.
Signed-off-by: Panagiotis Vasilikos <panagiotis.vasilikos@alexandra.dk>


ralight:
Thank you!

PR #1583, opened 2020-01-31
support bridge local clean session

karlp:
This is allows separating bridge clean session settings between the
local and remote endpoints.  Some broker implementations refuse to allow
non-clean sessions, as they don't support storing messages to be sent to
the connecting broker.  However, this doesn't meant that the local
broker can't be queueing messages to send _out_ to the remote broker.

This PR adds a new bridge connection setting, local_clean_session, that
allows controlling this split.

All ready now (I think)

- [] Have you successfully run `make test` with your changes locally?
no, but the test failure doesn't appear related to my code change.
and tests are added for the new features

-----

### DIFF
--- a/man/mosquitto.conf.5.xml
+++ b/man/mosquitto.conf.5.xml
@@ -1555,6 +1555,17 @@ openssl dhparam -out dhparam.pem 2048</programlisting>
 				</listitem>
 			</varlistentry>
 			<varlistentry>
+				<term><option>local_cleansession</option> [ true | false]</term>
+				<listitem>
+					<para>The regular <option>cleansession</option> covers both the local subscriptions
+					and the remote subscriptions.   local_cleansession allows splitting this.
+					Setting <replaceable>false</replaceable> will mean that the local connection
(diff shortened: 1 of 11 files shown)


ralight:
The idea seems sound to me.

In terms of the tristate bool, the `allow_anonymous` option is an `int8_t` that is initialised to -1, and is parsed as a bool in conf.c so we can tell between true/false/not set. I imagine something similar would work, along with doing the assignment `local_cleansession = cleansession` as appropriate in `conf.c:config__check()`

karlp:
Ok, I'll get back to and clean up the rest of it then.  Thanks for validating the idea :)

karlp:
blah, noticed some debug left in, will update again

karlp:
I've attempted to speed up the tests by interleaving the stop/starts more fully, but have been unable to get that to behave stably, so would rather get this in with a slow, but functional test.

ralight:
Merged, but with another commit to fix up the leaks :)

karlp:
I just realized, I never added this to the changelog?

PR #1599, opened 2020-02-15
CMakeLists.txt: Also install mqtt_protocol.h header file, mimic the M…

LocutusOfBorg:
…akefile

This fixes bugs such as:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=951116


karlp:
wouldnt' have been a problem if you were using the makefiles for linux, as recommended would it :)

LocutusOfBorg:
@karlp thanks to the switch, I discovered a serious bug (for Debian standards at least), in Makefiles too :)

https://github.com/eclipse/mosquitto/commit/7a5c2d4da514e8729ed52b9c48b09c04e5dfb469

We were using embedded library even if explicitly disabled... Having a cmake build failure helped me discover that the Debian version was outdated, while the package always built fine with the embedded library.

I still think having two build systems is not an ideal solution, but unless you want to get rid of cmake entirely, fixing it seems the best thing to do :)

LocutusOfBorg:
Btw the file has been never installed in Debian, even when it was Makefile only :)

LocutusOfBorg:
@ralight ping please? :)

ralight:
This has been fixed on the develop branch as part of other changes, which will be the next release anyway. Thanks for your report about this.

PR #1611, opened 2020-02-25
Feature/add deny option for acl

BrandtHill:
This features adds an option to the acl_file to allow a user to be explicitly denied access to a topic that might otherwise be granted from a broader topic. 
For example in an acl file:
```
user bob
topic readwrite api/#
topic deny api/sensitive/#
```
The user `bob` would be granted read/write access to all topics matching `api/#` with the exception of topics matching `api/sensitive/#`.
This allows us to configure mosquitto (no extra plugins) more easily without the need for extensive whitelists like this:
```
user bob
topic readwrite api/fun/#
topic readwrite api/stuff/#
topic readwrite api/hello/#
topic readwrite api/so/#
topic readwrite api/many/#
topic readwrite api/topics/#
...
```

Because tests aren't passing on `develop` branch currently, I also made these changes off `master` to test.

I hope the purpose of these changes was made clear.

Brandt

---
-----

### DIFF
--- a/man/mosquitto.conf.5.xml
+++ b/man/mosquitto.conf.5.xml
@@ -107,14 +107,16 @@
 						listed will have access. Topic access is added with
 						lines of the format:</para>
 
-					<para><code>topic [read|write|readwrite] &lt;topic&gt;</code></para>
+					<para><code>topic [read|write|readwrite|deny] &lt;topic&gt;</code></para>
 
-					<para>The access type is controlled using "read", "write" or
-						"readwrite". This parameter is optional (unless
+					<para>The access type is controlled using "read", "write",
+						"readwrite" or "deny". This parameter is optional (unless
 						&lt;topic&gt; includes a space character) - if not
 						given then the access is read/write.  &lt;topic&gt; can
 						contain the + or # wildcards as in
-						subscriptions.</para>
+						subscriptions. The "deny" option can used to explicity
+						deny access to a topic that would otherwise be granted
+						by a broader read/write/readwrite statement.</para>
 
 					<para>The first set of topics are applied to anonymous
 						clients, assuming <option>allow_anonymous</option> is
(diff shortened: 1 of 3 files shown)


BrandtHill:
Let me know if this feature seems reasonable and if any changes are required.

ralight:
Thank you for the nudge, and sorry for not getting to look at this earlier - I have been working on other things. I am definitely interested in this and will take a proper look soon.

dbeinder:
Looks like a neat feature!
One minor thing: With this change, the ACL check has to loop though all client ACLs, and exiting early on `SUCCESS` is no longer possible. If all `deny` ACLs were put into a separate deny list in `mosquitto__acl_user` and checked first, we could keep early exit and this could be implemented with no performance impact on existing configs.

ralight:
I think that's a good idea, ACL checking really has to be fast.

Sorry to be a pain, each of the commits has to have a signed-off-by line in them. Could you fix that please? I think it might be easiest for you to squash all the commits together and do that at the same time.

BrandtHill:
I'll work on adding a separate deny list for an early exit, definitely necessary for performance with large ACLs.

I will try to squash once changes are made.

BrandtHill:
I've changed it to add deny acl entries to the top of the list, so that once any denials have been passed during the check, any read|write matches can exit early with the assurance that it won't be caught by lingering denials. I've found this approach to be minimally invasive compared to adding a separate `mosquitto__acl*` for each user and one for patterns. It passes tests when based on the current `develop` branch with `make test WITH_CJSON=no`.

I've attempted to squash all of the commits into one with a signoff, but I'm not sure it had the desired effect.

BrandtHill:
Let me know if my solution looks good. It should have no performance impact on existing ACLs. 👍

dsb3:
+1 - this feature will help my use case as well.

ralight:
Thanks @BrandtHill, and sorry for the delay getting to it. It looks nice and is now merged.

BrandtHill:
Thank you so much, and I hope people find it useful and look forward to its release!

PR #1598, opened 2020-02-14
mosquitto.service: mosquitto starts even if the network is offline

svagionitis:
This patch changes the behavior of how mosquitto is starting using
systemd.

Currently it is starting only when the network is online, meaning that the
network is configured to a routable IP address.

With this patch, mosquitto is starting when the network is offline, does
not need to be configured. This is needed because the MQTT broker might
be used as an internal message bus which does not need the network to be
online.

Signed-off-by: Stavros Vagionitis <stavros.vagionitis@jci.com>

- [x] If you are contributing a new feature, is your work based off the develop branch?
- [ ] If you are contributing a bugfix, is your work based off the fixes branch?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Have you successfully run `make test` with your changes locally?
- [x] Have you signed the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php), using the same email address as you used in your commits?
- [x] Do each of your commits have a "Signed-off-by" line, with the correct email address? Use "git commit -s" to generate this line for you.

-----

PR #1617, opened 2020-03-05
Change systemd dependency to network.target

terokinnunen:
According to systemd "Running Services After the Network is up"
documentation [1], "network-online.target should not be pulled in
too liberally". Mosquitto can work perfectly well in local context
before network being online, hence network.target should be more
appropriate than network-online.target.

[1] https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

---
In our use case the network cable may even be unplugged and mosquitto
is used locally.

-----


karlp:
Duplicate of https://github.com/eclipse/mosquitto/pull/1598

PR #1629, opened 2020-03-15
1.deal with pubrel, ignore the ret of msg remove, 2.delete the same m…

RedCrab1:
- in function handle__pubrel(), ingore the return value of message__remove() . when deal with the   pubrel, received unexpected rel is apparently normal behaviour, if returned the rc, the loop may exit.
- in function message__queue(), add the check for same mid. if found the same mid, delete it. in some test scenarios, the server may reset, but client not reset , it may keep the older message, and the the next time connect to the server, it will received the same mid message.and then when get message from the queue, it will get unexpected message.

-----


ralight:
I believe this has already been fixed in a different way, so I'm closing this PR. Thanks for taking the time to submit it.

PR #1658, opened 2020-04-20
Client library publish helper

mgdm:

-----

### DIFF
--- a/lib/helpers.c
+++ b/lib/helpers.c
@@ -59,9 +59,7 @@ static void on_connect_publish(struct mosquitto *mosq, void *obj, int rc)
 	int result;
 	int i;
 
-    fprintf(stderr, "Got %d messages to publish\n", userdata->msg_count);
 	for (i=0; i < userdata->msg_count; i++){
-		fprintf(stderr, "Publishing message %d to topic %s\n", i, userdata->messages[i]->topic);
 		current_msg = userdata->messages[i];
 
 		result = mosquitto_publish(
--- a/lib/helpers.c
+++ b/lib/helpers.c
@@ -72,7 +72,7 @@ static void on_connect_publish(struct mosquitto *mosq, void *obj, int rc)
 		}
 	}
 
-    mosquitto_disconnect(mosq);
+	mosquitto_disconnect(mosq);
 }
 
 static void on_message_callback(struct mosquitto *mosq, void *obj, const struct mosquitto_message *message)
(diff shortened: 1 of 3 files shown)


lib/linker.version
```diff
@@ -83,6 +83,8 @@ MOSQ_1.5 {
 	global:
 		mosquitto_subscribe_simple;
 		mosquitto_subscribe_callback;
+		mosquitto_publish_single;
+		mosquitto_publish_multiple;
```
ralight, line 87: Could you please add these to the MOSQ_1.7 section at the bottom? That indicates they were added in version 1.7.

lib/mosquitto.h
```diff
@@ -2567,6 +2567,91 @@ libmosq_EXPORT int mosquitto_subscribe_callback(
...
+ *          no will.
+ *   tls - a libmosquitto_tls struct containing TLS related parameters, or NULL
+ *         for no use of TLS.
+ *
+ * Returns:
+ *   MOSQ_ERR_SUCCESS - on success
+ *   >0 - on error.
+ */
+libmosq_EXPORT int mosquitto_publish_multiple(
+		struct mosquitto_message **messages,
```
ralight, line 2599: Who owns **messages after this is called? It seems like the function doesn't modify the messages, so they should be declared as `struct mosquitto_message *const *const ` please.

lib/helpers.c
```diff
@@ -46,6 +52,28 @@ static void on_connect(struct mosquitto *mosq, void *obj, int rc)
...
+				mosq, NULL, current_msg->topic,
+				current_msg->payloadlen, current_msg->payload,
+				current_msg->qos, current_msg->retain);
+
+		if(result) {
+			break;
+		}
+	}
+
+	mosquitto_disconnect(mosq);
```
ralight, line 75: I'm not sure I like all of this happening in the connect callback, in particular the disconnect call. Could you change it so that the disconnect is called only after the messages are published? Maybe keeping a count of published messages in an on_publish() callback would be the easiest way, then calling disconnect there once the count is reached.

ralight (changes requested):
This looks good, thank you very much. I've added some specific comments on the review. I know I've mentioned adding support for MQTT v5 properties, but I'm inclined for the moment to accept this as it is and look at a v5 solution for both subscribers and publishers together.

Do you think you could put together some tests of the general publishing functionality? I can provide assistance if need be.

ralight:
This is now obsolete with the updated PR, so I'm closing it.

PR #1683, opened 2020-05-15
Feature: Publish Multiple Buffers

sidhant007:
Signed-off-by: sidhant007 <sidhbansal@gmail.com>

Hi, this is to support publishing multiple segments together, for data segments at different memory locations. Calling `mosquitto_publish` multiple times to send multiple segments is slower than the way described in this PR (which is to just do multiple `packet__write_bytes` for all the different segments) 

I have made the original `publish` a special case of the `publishm`. I could get the old tests passed when I made these changes on the master branch. I haven't yet been able to get the develop branch building on my pc, but I expect nothing should break.

If this approach to implement publishing multiple segments sounds good, then I can start working on designing the tests for this.


ralight:
What you are proposing is a new publishing function which allows a message payload to be constructed from multiple independent buffers, concatenated together.

In the case of a QoS 0 message, the buffers are passed straight to send__publishm(), which immediately copies them to the outgoing packet.

For QoS 1 and 2, the buffers are copied to the pending message payload, ready to be copied to the packet later on.

In both cases there isn't any change in the amount of memory copies going on, so the potential advantage lies in being able to make a single publish call without the user application having to create its own buffer.

From what you say, you are trying to avoid doing this:

```
for(i=0; i<buffer_count; i++){
    mosquitto_publish(..., buffer_len[i], buffers[i], ...);
}
```

This is a different outcome to what this pull request produces - multiple publishes versus a single publish. I think the real difference is between these two implementations:

With publishm:
```
buffer_count = sensor_count+2;
buffers = malloc(buffer_count * sizeof(char *));
buffer[0] = header;
for(i=0; i<sensor_count; i++){
    buffer[i+1] = sensor[i];
}
buffer[i+1] = footer;

buffer_lengths = malloc(buffer_count * sizeof(int));
for(i=0; i<buffer_count; i++){
    buffer_lengths[i] = strlen(buffer[i]);
}
mosquitto_publishm(..., buffer_lengths, buffers, buffer_count, ...);
free(buffers);
free(buffer_lengths);
```

Without publishm:
```
payload_len = strlen(header)+strlen(footer);
for(i=0; i<sensor_count, i++){
    payload_len += strlen(sensor[i]);
}
payload = malloc(payloadlen + 1);
memcpy(payload, header, strlen(header));
offset += strlen(header);
for(i=0; i<sensor_count; i++){
    len = strlen(sensor[i]);
    memcpy(&payload[offset], sensor[i], len);
    offset += len;
}
memcpy(&payload[offset], footer, strlen(footer));
mosquitto_publish(..., payload_len, payload, ...);
free(payload);
```

These are both the worst case situations, you may be able to get away with allocating buffers, buffer_lengths, and payload on the stack depending on your situation.

A better implementation might use an array of structs for passing the different payload buffers, as in `writev`:

```
struct iovec {
    void *iov_base;
    size_t iov_len;
}

buffer_count = sensor_count+2;
buffers = malloc(buffer_count * sizeof(struct iovec));
buffer[0]->iov_base = header;
buffer[0]->iov_len = strlen(header);
for(i=0; i<sensor_count; i++){
    buffer[i+1]->iov_base = sensor[i];
    buffer[i+1]->iov_len = strlen(sensor[i]);
}
buffer[i+1]->iov_base = footer;
buffer[i+1]->iov_len = strlen(footer);

mosquitto_publishv(..., buffers, buffer_count, ...);
free(buffers);
```

There is always going to be some overhead dealing with this situation, it just depends how we want to deal with that.

sidhant007:
Yes, @ralight your understanding about the purpose of `publishm` is correct. 

> From what you say, you are trying to avoid doing this: [code snippet you showed]

Yes, I am trying to avoid this and you are correct that the main advantage of this is so that the user can avoid creating their own buffer, i.e saving cost of calling `memcpy` on their own end.

I can modify the PR to implement `publishv` (instead of `publishm`) using iovec/iovec like struct. In our use case we are internally using an `iovec` so I would be more than glad to refactor this to use iovec/iovec-like struct directly. 

If we use iovec struct directly what do we want to do for platforms which don't support iovec? Using iovec will also create a concern on how does publish as a special case of publishv work for unsupported platforms? One way to handle is this is to duplicate the code of `send__publish` into `send__publishv`, another way would be to avoid using iovec and instead use an iovec-like struct of our own which will make `publishm` platform independent.

Let me know your thoughts about the iovec platform dependancy + code duplication between `send__publish` and `send__publishv` related concerns mentioned above and I will refactor this PR accordingly.

sidhant007:
Hey @ralight could you kindly let me know your thought about the questions I asked above.

So that I can take this PR in the intended direction and get it merged in.

My primary questions being: 
- Do we want to provide iovec-equivalent support for OS's/platforms that don't have iovec.
- Do you have any preference between the following two implementation styles for this PR:
      - making `send__publish` a special case of `send__publishv`
      - creating `send__publishv` independent of `send__publish`, will introduce some duplication of code at the benefit that there will be no coupling, so `send__publish` function will work just as it is for OS's/platforms that don't support iovec.

PR #1691, opened 2020-05-18
add a refcount to library init/cleanup

martingkelly:
Add a refcount around mosquitto_lib_init and mosquitto_lib_cleanup so that multiple calls to init/cleanup don't trigger memory leaks or double-frees.

Another for this change is the following scenario that I have:
- I have a library that internally uses mosquitto to subscribe to notifications.
- Applications want to use mosquitto to publish notifications.
- If both my library and applications call mosquitto init/cleanup, valgrind reports a memory leak, as the memory is clobbered in the second init call.

The only current solution is that the library requires the application to call mosquitto init/cleanup prior to initializing the library. However, it would be nice if instead both application and library could call init/cleanup and things would "just work" no matter what, especially as many applications won't use mosquitto and shouldn't have to bother about this implementation detail. Further, this breaks my library's abstraction in that if my library ever implements MQTT not using mosquitto (e.g. paho), the applications will be initializing mosquitto for no reason. In other words, this implementation detail (my library using mosquitto internally) should not be leaked to the user.

"make test" is failing on my machine, but it fails identically on the fixes branch (on `./02-subscribe-qos1.py`).
```
./02-subscribe-qos1.py c/02-subscribe-qos1-async2.test
connect_async failed: Connection refused
Traceback (most recent call last):
  File "./02-subscribe-qos1.py", line 47, in <module>
    (conn, address) = sock.accept()
  File "/usr/lib/python3.8/socket.py", line 292, in accept
    fd, addr = self._accept()
socket.timeout: timed out
make[2]: *** [Makefile:36: c] Error 1
make[2]: Leaving directory '/home/martin/ws/mosquitto/test/lib'
make[1]: *** [Makefile:11: test] Error 2
make[1]: Leaving directory '/home/martin/ws/mosquitto/test'
make: *** [Makefile:75: test] Error 2
```

-----

### DIFF
--- a/lib/mosquitto.c
+++ b/lib/mosquitto.c
@@ -34,6 +34,7 @@ Contributors:
 #include "packet_mosq.h"
 #include "will_mosq.h"
 
+static unsigned int init_refcount = 0;
 
 void mosquitto__destroy(struct mosquitto *mosq);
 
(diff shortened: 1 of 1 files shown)


karlp:
Are you sure you want to ignore the cleanup until the last refcount? (also, in general, you don't want to make unrelated whitespace changes)

martingkelly:
> Are you sure you want to ignore the cleanup until the last refcount?

How else would you suggest to do it? If the refcount is greater than 0, someone is still using the library, so it's not safe to cleanup the library.

> (also, in general, you don't want to make unrelated whitespace changes)

I agree, and tried not to make any. Where did I make these?

karlp:
Hrm, I guess, I was thinking if someone wanted to close it, they wanted it closed, but I guess that's not really useful.  Nevermind the whitespace comment, I was lost in the #ifdefs.

ralight:
Thanks for this, it's a good improvement.

Thinking about it now, I may go a bit further in the develop branch - changing mosquitto_new() to call lib init, and register lib cleanup to get called with atexit. With your changes, that would remove any requirement to call lib init/cleanup manually, but still allow a user to completely clean up if required.

martingkelly:
> Thinking about it now, I may go a bit further in the develop branch - changing mosquitto_new() to call lib init, and register lib cleanup to get called with atexit. With your changes, that would remove any requirement to call lib init/cleanup manually, but still allow a user to completely clean up if required.

That sounds good to me!

PR #1692, opened 2020-05-19
lazy init SSL for libmosquitto

abiliojr:
Signed-off-by: Abilio Marques <abiliojr@gmail.com>

Currently, when using a WITH_TLS compiled version of the mosquitto library, it will automatically initialize openssl as mosquitto_lib_init gets called. This causes > 120 Kb of RAM to be allocated, even if not needed (e.g., connecting to local broker, via TCP or UNIX sockets).

This patch allows for SSL to be initialized when it's needed. Broker functionality remains unchanged.

-------------------------------------------------------------------------------------------------------------

-----

### DIFF
--- a/lib/net_mosq.c
+++ b/lib/net_mosq.c
@@ -79,6 +79,8 @@ Contributors:
 int tls_ex_index_mosq = -1;
 UI_METHOD *_ui_method = NULL;
 
+static bool is_tls_initialized = false;
+
 /* Functions taken from OpenSSL s_server/s_client */
 static int ui_open(UI *ui)
 {
@@ -121,6 +123,7 @@ UI_METHOD *net__get_ui_method(void)
 {
 	return _ui_method;
 }
+
 #endif
 
 int net__init(void)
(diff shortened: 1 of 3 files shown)


lib/net_mosq.c
```diff
@@ -537,6 +544,10 @@ int net__socket_connect_tls(struct mosquitto *mosq)
 {
 	int ret, err;
 
+    if (!is_tls_initialized) {
+	    net__init_tls_engine();
+    }
+
```
ralight, line 550: Could you please replace spaces here with a tab.

lib/net_mosq.c
```diff
@@ -184,6 +170,27 @@ void net__cleanup(void)
 #endif
 }
 
+#ifdef WITH_TLS
+void net__init_tls_engine(void)
```
ralight, line 174: "engine" has a specific meaning in the context of openssl. Could you please rename this to be `net__init_tls()` because it is doing general tls init rather than just the engine.

lib/net_mosq.c
abiliojr: oops... changed

lib/net_mosq.c
abiliojr: done

ralight:
I've been testing this, and I'm afraid it doesn't work - quite a few of the client library tests fail. It also always re-initialises TLS regardless of whether it has already been done.

abiliojr:
> I've been testing this, and I'm afraid it doesn't work - quite a few of the client library tests fail. It also always re-initialises TLS regardless of whether it has already been done.

Should be fixed by now... tls initialization was happening too late, and that was detected by the failing tests. It should not longer re-initialize tls.

Squashed commits for cleanliness.

ralight:
Thanks for making the changes, that's now merged.

PR #1706, opened 2020-05-26
Fix Bug for message out-going

zero0714:
1 database.c 1033:When test message subscribing, we found that sometimes the broker would not push the message if there was a message expired(there would be no more message be put into inflight quota), so increase send quota when removing the expired message.

2 database.c 765:"send quota" take the place of "send quota" 
-----
- [N/A] Have you signed the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php), using the same email address as you used in your commits?
- [N/A] If you are contributing a new feature, is your work based off the develop branch?


ralight:
Thanks for the PR, this has been fixed independently of it so I'm going to close it

PR #1736, opened 2020-07-02
Feature: Add support for publishing multiple buffers concatenated together

sidhant007:
As discussed in my previous [PR](https://github.com/eclipse/mosquitto/pull/1683) (which I broke accidentally while force-pushing an unintended commit and subsequently closed this PR)

Motivation: If you want to make a single publish to consist of multiple buffers concatenated together, then instead of concatenating the buffers on the user's end and then passing it to `mosquitto_publish` they can call use `mosquitto_publish_bufs`, this will reduce `memcpy` calls on the user's end. 
Example - If the user has 10 buffers each 20 bytes big, then the user will save cost (in terms of both, time and memory) of doing `memcpy` for a total of 200 bytes.

Signature of the function: `int mosquitto_publish_bufs(mosq, mid, topic, buffers, buffers_cnt, qos, retain)`, where `buffers` is an array of struct `buf`, where
```
struct buf {
   const void *base;
   uint32_t len;
};
```
^Taking into  account [the feedback](https://github.com/eclipse/mosquitto/pull/1683#issuecomment-631438763) provided on the last PR.

The difference in implementations for users will look like following:

With publish_bufs: 
```
buffer_count = sensor_count + 2;
buffers = malloc(buffer_count * sizeof(struct buf));
buffer[0]->iov_base = header;
buffer[0]->iov_len = strlen(header);
for(i=0; i<sensor_count; i++){
    buffer[i+1]->base = sensor[i];
    buffer[i+1]->len = strlen(sensor[i]);
}
for(i=0;i<4;i++){
	buffers[i].base = message[i];
	buffers[i].len = strlen(message[i]);
}
mosquitto_publish_bufs(..., buffers, buffer_count, ...);
free(buffers);
```

Without publish_bufs: 
```
payload_len = strlen(header)+strlen(footer);
for(i=0; i<sensor_count, i++){
	payload_len += strlen(sensor[i]);
}
payload = malloc(payloadlen + 1);
memcpy(payload, header, strlen(header));
offset += strlen(header);
for(i=0; i<sensor_count; i++){
	len = strlen(sensor[i]);
	memcpy(&payload[offset], sensor[i], len);
	offset += len;
}
memcpy(&payload[offset], footer, strlen(footer));
mosquitto_publish(..., payload_len, payload, ...);
free(payload);
```

- I have defined a new `struct buf` instead of using `struct iovec` (present in  <sys/uio.h>) because of the feedback provided [here](https://www.eclipse.org/lists/mosquitto-dev/msg02359.html) of keeping the mosquitto API as consistent as possible across platforms.
- The implementation makes `mosquitto_publish` a special case of `mosquitto_publish_bufs` to reduce code-duplication. I can make the two separate in-case the maintainer prefers so.
- The three new tests added (03-publish_bufs-qos0, 03-publish_bufs-qos0-no-payload, 03-publish_bufs-c2b-qos1-disconnect) extend the code-coverage to the new functionality added (primarily when `buffers_cnt != 1`) and the currently present (03-publish-*) tests still pass confirming that `mosquitto_publish_bufs` works fine when `buffers_cnt == 1`.

-----

### DIFF
--- a/lib/actions.c
+++ b/lib/actions.c
@@ -34,8 +34,27 @@ int mosquitto_publish(struct mosquitto *mosq, int *mid, const char *topic, int p
 	return mosquitto_publish_v5(mosq, mid, topic, payloadlen, payload, qos, retain, NULL);
 }
 
+int mosquitto_publish_bufs_v5(struct mosquitto *mosq, int *mid, const char *topic, const struct buf *buffers, int buffers_cnt, int qos, bool retain, const mosquitto_property *properties);
+
+int mosquitto_publish_bufs(struct mosquitto *mosq, int *mid, const char *topic, const struct buf *buffers, int buffers_cnt, int qos, bool retain)
+{
+  return mosquitto_publish_bufs_v5(mosq, mid, topic, buffers, buffers_cnt, qos, retain, NULL);
+}
+
 int mosquitto_publish_v5(struct mosquitto *mosq, int *mid, const char *topic, int payloadlen, const void *payload, int qos, bool retain, const mosquitto_property *properties)
 {
+  const struct buf buffer = {payload, (uint32_t) payloadlen};
+  return mosquitto_publish_bufs_v5(mosq, mid, topic, &buffer, 1, qos, retain, properties);
+}
+
+int mosquitto_publish_bufs_v5(struct mosquitto *mosq, int *mid, const char *topic, const struct buf *buffers, int buffers_cnt, int qos, bool retain, const mosquitto_property *properties)
+{
+    int i;
+    int payloadlen = 0;
+    int offset = 0;
+    for(i=0; i<buffers_cnt; i++){
+        payloadlen += (int)buffers[i].len;
+    }
 	struct mosquitto_message_all *message;
 	uint16_t local_mid;
 	const mosquitto_property *p;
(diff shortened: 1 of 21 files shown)


sidhant007:
Update: The latest changes on the develop branch (unrelated to this PR), break the tests.

`make ptest -WITH_CJSON=no` gives this:
```
persist_read_stubs.c:48:5: error: conflicting types for ‘db__message_store’
 int db__message_store(struct mosquitto_db *db, const struct mosquitto *source, struct mosquitto_msg_store *temp, struct mosquitto_msg_store **stored, uint32_t message_expiry_interval, dbid_t store_id, enum mosquitto_msg_origin origin)
     ^~~~~~~~~~~~~~~~~
In file included from persist_read_stubs.c:7:0:
../../src/mosquitto_broker_internal.h:661:5: note: previous declaration of ‘db__message_store’ was here
 int db__message_store(struct mosquitto_db *db, const struct mosquitto *source, struct mosquitto_msg_store *stored, uint32_t message_expiry_interval, dbid_t store_id, enum mosquitto_msg_origin origin);
     ^~~~~~~~~~~~~~~~~
```

Also @ralight, please let me know if you have some feedback for this PR.

zadewg:
I am interested in this PR too. Can we have a review?

Issue #1659, opened 2020-04-21
make `make test` work under CentOS8/python3


honglei:
`make test` output:
```
cc -I../../../lib -I../../../src -Wall -Werror -fPIC -shared auth_plugin_msg_params.c -o auth_plugin_msg_params.so
cc -I../../../lib -I../../../src -Wall -Werror -fPIC -shared auth_plugin_extended_multiple.c -o auth_plugin_extended_multiple.so
cc -I../../../lib -I../../../src -Wall -Werror -fPIC -shared auth_plugin_extended_single.c -o auth_plugin_extended_single.so
cc -I../../../lib -I../../../src -Wall -Werror -fPIC -shared auth_plugin_extended_single2.c -o auth_plugin_extended_single2.so
cc -I../../../lib -I../../../src -Wall -Werror 08-tls-psk-pub.c -o 08-tls-psk-pub.test ../../../lib/libmosquitto.so.1
cc -I../../../lib -I../../../src -Wall -Werror 08-tls-psk-bridge.c -o 08-tls-psk-bridge.test ../../../lib/libmosquitto.so.1
make[3]: Leaving directory '/root/libs/mosquitto-1.6.9/test/broker/c'
./01-connect-anon-denied.py
Traceback (most recent call last):
  File "./01-connect-anon-denied.py", line 22, in <module>
    broker = mosq_test.start_broker(filename=os.path.basename(__file__), use_conf=True, port=port)
  File "/root/libs/mosquitto-1.6.9/test/mosq_test.py", line 63, in start_broker
    raise IOError
OSError
make[2]: *** [Makefile:23: 01] Error 1
make[2]: Leaving directory '/root/libs/mosquitto-1.6.9/test/broker'
make[1]: *** [Makefile:10: test] Error 2
make[1]: Leaving directory '/root/libs/mosquitto-1.6.9/test'
make: *** [Makefile:75: test] Error 2
```

```
[root@localhost broker]# ../../src/mosquitto -v -c 01-connect-anon-denied.conf
1587476916: mosquitto version 1.6.9 starting
1587476916: Config loaded from 01-connect-anon-denied.conf.
1587476916: Opening ipv4 listen socket on port 1888.
1587476916: Opening ipv6 listen socket on port 1888.
1587476916: Error: Invalid user 'mosquitto'.
```
So, I add user `mosquitto`, and rerun it:

```
make[3]: Leaving directory '/root/libs/mosquitto-1.6.9/test/lib/cpp'
[...]

PR #1745, opened 2020-07-10
Fix IPv6 socket for Windows.

SigmundVik:
The IPV6_V6ONLY call to setsockopt() in net__socket_listen()
contained two mistakes for Windows.

The first mistake was that the type of "ss_opt" was wrong.
From the setsockopt() Win32 API documentation: "The optlen
parameter should be equal to sizeof(int) for Boolean options."
https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-setsockopt#remarks

The second mistake was that the IPV6_V6ONLY flag must be flipped
on Windows, see for instance how Boost.Asio does this:
https://github.com/boostorg/asio/blob/2d27fc712387c54b7eb45347fc4669be29a38d9e/include/boost/asio/detail/impl/socket_ops.ipp#L1806

This commit also changes the code to only set the IPV6_V6ONLY
flag for IPv6 (similarly to how Boost.Asio does it), because
that seems more correct.

Before this commit, starting the Mosquitto broker when the IPv6 port
already was in use would produce an output like this (and the broker
would be running):

1594385168: mosquitto version 1.6.9 starting
1594385168: Using default config.
1594385168: Opening ipv6 listen socket on port 31001.
1594385168: Opening ipv4 listen socket on port 31001.

With this commit, starting the Mosquitto broker when the IPv6 port
already is in use makes the broker correctly terminate with an exit
code of 1 and this output:

1594385193: mosquitto version 1.6.9 starting
1594385193: Using default config.
1594385193: Opening ipv6 listen socket on port 31001.
1594385193: Error: Only one usage of each socket address (protocol/network address/port) is normally permitted.

Signed-off-by: Sigmund Vik <sigmund_vik@yahoo.com>

-----

(Just like before this commit, I need to apply a [hack](https://github.com/eclipse/mosquitto/issues/1659#issuecomment-653629126) to `test/lib/02-subscribe-qos1.py` in order to make all tests pass on my Ubuntu 18.04 machine.)

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -588,11 +588,7 @@ int net__socket_listen(struct mosquitto__listener *listener)
 	struct addrinfo *ainfo, *rp;
 	char service[10];
 	int rc;
-#ifndef WIN32
 	int ss_opt = 1;
-#else
-	char ss_opt = 1;
-#endif
 #ifdef SO_BINDTODEVICE
 	struct ifreq ifr;
 #endif
@@ -646,8 +642,14 @@ int net__socket_listen(struct mosquitto__listener *listener)
 		setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, &ss_opt, sizeof(ss_opt));
 #endif
 #ifdef IPV6_V6ONLY
-		ss_opt = 1;
-		setsockopt(sock, IPPROTO_IPV6, IPV6_V6ONLY, &ss_opt, sizeof(ss_opt));
+		if(rp->ai_family == AF_INET6){
+#ifndef WIN32
+			ss_opt = 1;
+#else
+			ss_opt = 0;
+#endif
+			setsockopt(sock, IPPROTO_IPV6, IPV6_V6ONLY, &ss_opt, sizeof(ss_opt));
+		}
 #endif
 
 		if(net__socket_nonblock(&sock)){


ralight:
I agree that ss_opt should be an int for Windows, thanks for pointing that out.

The other part of what you are saying does not match the behaviour that I see. I have tested on Windows 10 only, but it works as expected by producing an error if trying to open an IPv6 socket for a port where one is already open. In fact if it was possible to open a socket with the same port but without setting `SO_REUSEPORT` then there would be a serious problem.

On Windows my understanding that the default behaviour is that IPV6_V6ONLY is set anyway. I don't think it is a problem to set this option on IPv4 sockets, and so there is no need to introduce extra complexity.

If you'd like to change the commit to just make the char to int change I'll happily accept it, otherwise if you'd prefer I can just make the change myself.

SiggyBar:
Ah, when I run two programs that both set `IPV6_V6ONLY` to 1 (like two instances of the Mosquitto broker), then I do get the expected behavior (`Error: Only one usage of each socket address (protocol/network address/port) is normally permitted.`).  However, if the Mosquitto broker sets `IPV6_V6ONLY` to 1 and another program sets `IPV6_V6ONLY` to 0, then that does not work.  I agree that setting `IPV6_V6ONLY` to 1 the way Mosquitto does it is fine (I will update this PR to only include the `ss_opt` fix), but I am not sure how to deal with programs that sets `IPV6_V6ONLY` to 0 (which is the default behavior for `Boost.Asio`)...

> On Windows my understanding that the default behaviour is that IPV6_V6ONLY is set anyway. I don't think it is a problem to set this option on IPv4 sockets, and so there is no need to introduce extra complexity.

I agree that it is probably not a problem to set the `IPV6_V6ONLY` flag for `IPPROTO_IPV4`, it just looks a bit odd.

ralight:
That's very interesting. I wonder if the case you describe works because they are effectively binding to different interfaces.

SiggyBar:
The behavior in this regard seems to be different between Windows 10 and Ubuntu 18.04.  On Ubuntu, if I create one version of the Mosquitto broker with `IPV6_V6ONLY` set to 1 (and no IPv4) and another version with `IPV6_V6ONLY` set to 0, then they will not both run using the same port.  The one that is started last will produce the expected `Address already in use` error.  This is unlike Windows, where unfortunately both brokers will start up without problems.

I did a test on Windows where the Mosquitto broker sets up two sockets when `rp->ai_family == AF_INET6`, one with `IPV6_V6ONLY` set to 1 and another with `IPV6_V6ONLY` set to 0.  With this change the broker that is started last will produce the expected `Only one usage of each socket address (protocol/network address/port) is normally permitted.` error.  Is this a solution we want to persuade further?  If we are afraid that this change might have unforeseen consequences, then we could silently ignore if listening on the dual-stacked IPv6 socket fails (since the only purpose of doing this is to detect potential port conflicts).

> If you'd like to change the commit to just make the char to int change I'll happily accept it, otherwise if you'd prefer I can just make the change myself.

Contrary to what I indicated earlier, it would be great if you could make that change.  Then this PR can be used to discuss what to do about dual-stack sockets for IPv6 on Windows.

SiggyBar:
Here is the [commit](https://github.com/SiggyBar/mosquitto/commit/7ed57a1211983af974e9978ebb1a182c7e7a4460) with my suggested fix.

It also contains a fix that closes all opened sockets (instead of just the last one in some cases) when an error occurred.  This should be separated out into its own commit.

PR #1746, opened 2020-07-10
Add broker ready message.

SiggyBar:
Before this commit there was no good way to detect that the
Mosquitto broker was done with its startup phase on systems
without systemd.

On such systems it was tricky to e.g. start the broker from
a test where ports are dynamically assigned and one have to
deal with potential port conflicts.  Without a way to know
that the broker is done with its startup phase, there was no
way to know if the broker was able to acquire the port (for
both IPv4 and IPv6) without waiting for some unknown period
of time (when many tests are run in parallel a single process
might be starved for resources).

With this new broker ready message it is easy for the parent
process to monitor the broker output and figure out when the
port was successfully acquired.

Signed-off-by: Sigmund Vik <sigmund_vik@yahoo.com>

-----

### DIFF
--- a/src/mosquitto.c
+++ b/src/mosquitto.c
@@ -364,6 +364,7 @@ int main(int argc, char *argv[])
 	}
 #endif
 
+	log__printf(NULL, MOSQ_LOG_INFO, "mosquitto version %s running", VERSION);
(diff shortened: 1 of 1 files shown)


ralight:
That looks fine, thanks very much.

Issue #769, opened 2018-04-06
Keep persistent sessions after mosquitto restart


nixikanius:
I have a problem with losing of messages for persistence sessions which came after the mosquitto restarts until clients with persistence sessions reconnect. As for me, it seems that broker does not pay attention to sessions that existed before the mosquitto restart but wherein carefully saves all pending data to persistence db on shutdown.

So the problem scenario:

1. Start the mosquitto.
2. Start a client with persistence session, QoS 2 and subscription to some test topic.
3. Send test data to test topic.
3a. Test data successfully received by persistence client.
4. Stop persistence client.
5. Restart the mosquitto.
6. Send something else test data to test topic.
7. Start persistence client.
7a. **Test data from step 6 does not come to persistence client.**

The mosquitto key config parameters:

```
persistence true
persistence_file mosquitto.db
persistence_location /mosquitto/data/
autosave_interval 1
autosave_on_changes true
queue_qos0_messages true
```

Is there any way to solve the described problem?

karlp:
What version of mosquitto?  Are you sure you're using not just clean session=false, but also the same client id on both connections?  This is pretty core behaviour that people rely on day in, day out.

nixikanius:
I also think that case I described should be ordinal. I use mosquitto 1.4.12 in docker and tested mosquitto 1.4.15 too. As a persistent client, I use [telegraf](https://github.com/influxdata/telegraf) 1.5.3 [mqtt_consumer](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/mqtt_consumer) which pushes data to InfluxDB.

There are not any problems with messages sent before mosquitto restart:

1. Start the mosquitto.
2. Start a client with persistence session, QoS 2 and subscription to some test topic.
3. Send test data to test topic.
3a. Test data successfully received by persistence client.
4. Stop persistence client.
5. Send something else test data to test topic.
6. Restart the mosquitto.
7. Start persistence client.
7a. Test data from step 5 successfully received by persistence client.

Think, above scenario should sufficiently demonstrate that telegraf client sets `clean session = false` and uses the same client id for all connections. The only problem is to deliver messages which sent between mosquitto restart and persistence client reconnects.

ralight:
Can you look at the persistence file saved by mosquitto when you stop it? You can use https://github.com/eclipse/mosquitto/tree/master/src/db_dump to get a more human friendly output.

[...]

Issue #384, opened 2017-02-21
Broker fails to store QoS1 message for persistent client after restart


codami:
My setup:
Broker: with persistence enabled (mosquitto.db)
Client1: non persistent (publisher that follows the ACL pattern - see below)
Client2: persistent (subscriber that authenticates as "superuser")
ACL file:
```
user superuser
topic readwrite #

pattern readwrite %u/%c/#
```
Scenario:
1. Start Broker
2. Client2 connects as "superuser" and subscribes successfully for "+/+/status" topic with QoS2
3. Stop Broker
4. Start Broker
5. Client1 connects and publishes for "User1/Client1/status" topic a message with QoS1
6. Client2 connects and subscribes again as before

In this case Client2 should receive the message Client1 sent as soon as it subscribes, but it doesn't.

After some debugging I've got to the point where the ACL check fails. It never gets through [this loop](https://github.com/eclipse/mosquitto/blob/master/src/security_default.c#L245).
I believe the ACL check would have worked for a user subject to the ACL pattern, because these patterns are checked later in the [next loop](https://github.com/eclipse/mosquitto/blob/master/src/security_default.c#L266).

codami:
As suspected, new tests show that the ACL check fails only for user definitions and works fine for ACL patterns.

ralight:
The "problem" here is that persisted clients don't have the username they connected with saved to the persistence db (the rationale being that the messages are associated with a client id and the username may change on reconnection). That would need changing to fix this.

codami:
Following the same rationale would mean that in this scenario (without Broker restart):
1. Start Broker
2. Client2 connects and subscribes
3. Client2 disconnects
4. Client1 connects and publishes
5. Client 2 reconnects

the Broker should also not store any message for Client2 while disconnected.
In my tests it does store them and moreover it uses the username from Client2's previous connection.
Consequently Client2 might loose messages when it connects with a different username. 

For me it looks like the "ACL check" step happens too early in the publishing process.
It should be postponed until it has enough info to decide. The Broker could filter messages once Client2 connects and the username is known.

Thank you for looking into this issue!

codami:
[This](https://github.com/codami/mosquitto/commit/d82e5874caecaf071f238699fb784ba4c6b9b415) is a change that worked for me in both scenarios. Do you see any problems with this kind of solution?

ralight:
This should now be fixed in the fixes branch.

PR #1769, opened 2020-08-06
Fix lost messages after restart with persistent sessions

dbeinder:
Fixes https://github.com/eclipse/mosquitto/issues/769 / https://github.com/eclipse/mosquitto/issues/384 see repro in https://github.com/eclipse/mosquitto/issues/769#issuecomment-580281178 by @jr-dimedis
This stores the last connected username and listener along with persistent client sessions and attempts to reattach ACLs to the client context, so messages can be retained for clean=false sessions.

With `per_listener_settings ` off, reattaching the ACLs is straightforward, and the previous listener port is ignored. That way even when changing ports, no data is lost.
With `per_listener_settings` on, I don't think there is a 'correct' way to handle a changed port-config - I have adopted the same strategy used for persisting the message source listener, simply store the port and try to find matching listener. Only when changing listener ports in the config and restarting some subscriptions may lose their ACLs until the next subscriber connect. 

There was no unused padding in the client chunk, so I had to bump the DB version.
Do the database changes look good? Are there any unintended side-effects; @ralight suspected some in https://github.com/eclipse/mosquitto/issues/769?

For now this is based on master, since develop doesn't build on windows for me (some sigset_t dependency).
Or should I rebase onto fixes?
I haven't touched the tests yet, since they are based on database binaries.


ralight:
Thanks for looking into this. I think this would be best off based on `fixes`, despite it introducing a db format change. I don't think there are any other consequences of making this change.

In terms of the tests, duplicating `TEST_v5_client` as `TEST_v6_client`, and add [v6-client.txt](https://github.com/eclipse/mosquitto/files/5035086/v6-client.txt) as `v6-client.test-db` - I think I've got it right.

dbeinder:
Because the DB version is stored as `v6` in all write tests, and the write tests share binaries with the read tests, I had to update most v5 binaries. The client chunk is the same for all tests, I added username=`usrname` and listener_port=`1883` to all of them. I renamed most of the v5 tests to v6 and kept the v5 client test. The only really new test is `TEST_v6_client`.

I rebased on `fixes`, this now passes all tests on my machine. @ralight 

Unrelated to this PR: When debugging on the fixes branch, I had to add the calls to `fflush()` on `stdout/stderr` [[here](https://github.com/eclipse/mosquitto/commit/54f3b686dc2a52d952ccf03945c19985440eb49c#diff-f909d0b1a298531f76132943dc55f86aL296)], not sure if that was intended with the changes to file logging. In any case, `setvbuf` with `_IOLBF` is not supported on Windows, that should probably have some platform workaround.

ralight:
`_IOLBF`, urgh how tedious :+1: 

Thanks for the test updates, it looks good to me.

PR #1794, opened 2020-08-18
do not include pthread when compiling without threading support

titouanc:
This fixes the following error, when compiling for systems without
pthread support, and when passing WITH_THREADING=no to make:

    thread_mosq.c:24:12: fatal error: pthread.h: No such file or directory
     #  include <pthread.h>
                 ^~~~~~~~~~~
    compilation terminated.


ralight:
Ah, sorry about that. Thanks for the fix.

PR #1818, opened 2020-09-08
Added support for SSL keylogging.

petertorelli:
This feature adds the `-k` option, which installs a callback function to log SSL key material to the logfile.

-----

- [ X ] Have you signed the [Eclipse Contributor Agreement](https://www.eclipse.org/legal/ECA.php), using the same email address as you used in your commits?
- [ X ] Have you added an explanation of what your changes do and why you'd like us to include them?
- [  ] Have you successfully run `make test` with your changes locally?

-----

Couldn't run `make test` due to CUnit dependencies. Already committed and pushed so `commit -s` wasn't available, but if there is a way to add this after the fact, I will do so.

### DIFF
--- a/src/conf.c
+++ b/src/conf.c
@@ -239,6 +239,7 @@ void config__init(struct mosquitto_db *db, struct mosquitto__config *config)
 	config__init_reload(db, config);
 
 	config->daemon = false;
+	config->keylog = false;
 	memset(&config->default_listener, 0, sizeof(struct mosquitto__listener));
 	config->default_listener.max_connections = -1;
 	config->default_listener.protocol = mp_mqtt;
@@ -357,10 +358,11 @@ static void print_usage(void)
 {
 	printf("mosquitto version %s\n\n", VERSION);
 	printf("mosquitto is an MQTT v5.0/v3.1.1/v3.1 broker.\n\n");
-	printf("Usage: mosquitto [-c config_file] [-d] [-h] [-p port]\n\n");
+	printf("Usage: mosquitto [-c config_file] [-d] [-h] [-p port] [-k]\n\n");
 	printf(" -c : specify the broker config file.\n");
 	printf(" -d : put the broker into the background after starting.\n");
 	printf(" -h : display this help.\n");
+	printf(" -k : log SSL keys to logfile.\n");
 	printf(" -p : start the broker listening on the specified port.\n");
 	printf("      Not recommended in conjunction with the -c option.\n");
 	printf(" -v : verbose mode - enable all logging types. This overrides\n");
@@ -391,6 +393,8 @@ int config__parse_args(struct mosquitto_db *db, struct mosquitto__config *config
 		}else if(!strcmp(argv[i], "-h") || !strcmp(argv[i], "--help")){
 			print_usage();
 			return MOSQ_ERR_INVAL;
+		}else if(!strcmp(argv[i], "-k")){
+			config->keylog = true;
 		}else if(!strcmp(argv[i], "-p") || !strcmp(argv[i], "--port")){
 			if(i<argc-1){
 				port_tmp = atoi(argv[i+1]);
--- a/src/net.c
+++ b/src/net.c
@@ -307,6 +307,11 @@ static unsigned int psk_server_callback(SSL *ssl, const char *identity, unsigned
 }
 #endif
 
+void ssl_keylog_callback_func(const SSL *ssl, const char *line)
+{
+	log__printf(NULL, MOSQ_LOG_INFO, "SSL Keylog: %s", line);
+}
+
 #ifdef WITH_TLS
 int net__tls_server_ctx(struct mosquitto__listener *listener)
 {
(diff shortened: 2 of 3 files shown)


petertorelli:
I'll check into the correct branch and sign off.

src/net.c
```diff
@@ -307,13 +307,19 @@ static unsigned int psk_server_callback(SSL *ssl, const char *identity, unsigned
 }
 #endif
 
+void ssl_keylog_callback_func(const SSL *ssl, const char *line)
```
ralight, line 310: With a few exceptions, TLS is used instead of SSL throughout the project. Would you please change this to `tls_` and make similar changes in the few other places where SSL is used.

Please also guard this function inside a `#ifdef WITH_TLS` block and check it compiles when using `make WITH_TLS=no`.

src/net.c
```diff
@@ -307,13 +307,19 @@ static unsigned int psk_server_callback(SSL *ssl, const char *identity, unsigned
...
+{
+	log__printf(NULL, MOSQ_LOG_INFO, "SSL Keylog: %s", line);
```
ralight, line 312: Is there any value in adding the client address/port to the log line, to allow distinguishing between different clients? It makes sense to me, but I haven't used this sort of functionality before. If so then using `SSL_get_rbio()` and `BIO_get_fd()` to get the socket, and tweaking `net__socket_get_address()` to also return the port would allow you to change the message to `TLS Keylog for %s:%d %s`.

src/conf.c
```diff
@@ -391,6 +393,8 @@ int config__parse_args(struct mosquitto_db *db, struct mosquitto__config *config
 		}else if(!strcmp(argv[i], "-h") || !strcmp(argv[i], "--help")){
 			print_usage();
 			return MOSQ_ERR_INVAL;
+		}else if(!strcmp(argv[i], "-k")){
```
ralight, line 396: Where possible I prefer to keep short command line arguments for common, frequently used options. This is an option that should not be used in production, so I'd like something that is more effort to type. Could you please change to `--tls-keylogging`.

ralight (commented):
(I realise that you've closed this pending putting it on `develop`, but here is a review anyway).

I like this, it's a nice simple addition for making debugging with TLS easier. I've put some comments inline in the code.

ralight:
Thinking about this some more, I wonder if it would be better to log to a separate file. We'd have `--tls-keylog /path/to/file` as the option, and the callback would simply append to the file.

It probably wouldn't be necessary to include the address:port portion either, because we don't need to identify which parts of the log are relevant.

The advantage is that it follows the key log file format directly, and so can be used with anything that supports it without any messing with logs.

petertorelli:
@ralight Thanks for all the suggestions. I will implement them and do a proper pull request once I get local test build passing (still working on the docs issue from the other comment). Regarding the log messages' destination: the reason why I used the primary log was to make it more like systemd logging: everything in one file & grep what you need, or more simply, one-log-per-program; plus this is only for debug which assumes it only be used under rare circumstances. How about I keep it in the main log now and if there are complaints, I can move it later? Thanks.

ralight:
@petertorelli I register my complaint already :)

If we have a separate file just for the the key log, in the standard key log format that e.g. wireshark expects, you can tell wireshark to use that file and it will automatically load the new keys as clients connect and so decryption is entirely automatic. 

Keeping it in the main log would mean a separate script would be required to pull out that information. Why not save everybody the trouble?

petertorelli:
> e.g. wireshark expects,

We'll there's the problem: I didn't know of any application that read the OpenSSL key log directly, thanks for showing me. Also, you mentioned it should be opened in "append" mode. This brings up two questions: 1) is wireshark (& other apps) smart enough to use the tail of the key log file? 2) is logfile bloat a problem - or will people generally add a logrotated rule if necessary (or is the file too small to worry about)?

ralight:
1. In the quick test I did, wireshark picked up the additions to the file whilst it was running. 
2. I said use append, so that we're not keeping the file open constantly - which can be a problem on some systems. I think we've got a bit of a pass here because this really is only for debugging, so logfile bloat shouldn't be a problem. If someone is considering leaving this running long enough to need logrotation... they're probably doing something wrong. But either way, the open/close with append would let them do a rotation.

ralight:
@petertorelli If you can fix the commit message to include the signed-off-by line then I'll merge this and then massage it into the format that I'd like. You get the credit for the original work, and I get to put it in the next version, which I'd like.

Something like this should do the job:
```
git commit -s --amend
git push -f
```

petertorelli:
Hi Roger,
Sorry for the long delay, I hadn't been able to build the tests and then got I side-tracked with work. If I can't get it done this week I'll make the amendment.

Thanks,
Peter

ralight:
I've now implemented this separately, so I'm going to close this request. Thanks for the idea and direction.

petertorelli:
Hi Roger,
Thanks for integrating this PR. It is much appreciated.
Peter

PR #1882, opened 2020-11-09
feature: allow library to use ppool to avoid select limitations.

skolodzieski:
The inherent limit on the number of file descriptor available using the select() system call was preventing the use of the mosquitto library. loop.c was modified to allow the use of ppoll() to avoid such limitations. This change was not hooked into cmake, but is conditional on the HAVE_PPOLL pre-processor define.

### DIFF
--- a/lib/loop.c
+++ b/lib/loop.c
@@ -16,9 +16,20 @@ Contributors:
 
 #include "config.h"
 
+/*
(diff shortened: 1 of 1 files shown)


skolodzieski:
Added the Signed-off-by footer.

skolodzieski:
Cleaning up.

PR #1907, opened 2020-11-20
CMake: Make building the apps and plugins optional

Hallot:
Useful in case only the lib is needed.

Signed-off-by: Pierre Hallot <hallotpierre@gmail.com>


ralight:
This looks good, thanks for testing out the develop build!

PR #1920, opened 2020-12-01
CMake: Improve handling of cJSON dependency

Hallot:
Fix warning about name mismatch between the `add_package` command and the FindcJSON.cmake script.  
Require finding cJSON so that the build fails early at generation time instead of compile time in case it is not found.

-----


CMakeLists.txt
```diff
@@ -122,7 +122,7 @@ if (WITH_APPS)
 endif (WITH_APPS)
 
 if (WITH_PLUGINS)
-	FIND_PACKAGE(cJSON)
+	find_package(cJSON REQUIRED)
```
ralight, line 125: It isn't required for all plugins, so I don't think this is correct.

Hallot: You are right, my bad.  
Actually the issue is that some app/client can require this dependency (`mosquitto_ctrl`, `sub_client_output`), thus if plugins are disabled it can lead to issues.  
I have moved the `find_package` out of the plugin check, but kept it optional.

ralight:
Yes, that's much better. Thanks, I hadn't spotted that it would fail if plugins were disabled.

Issue #188, opened 2016-06-07
Mosquitto client implements QoS 1 a zero or more times and QoS 2 as at most once


fhars:
In both cases, an adversary scheduler will run the code right up to the `if(mosq->on_message){` and then reboot the system. When the clients reconnects with clean session equal to 0, the server will regard the message as completely acknowledged, while the application will never have seen the message at all.

This is related to #118, and there are several bad choices here. It may actually be impossible to implement QoS 2 correctly without writing the message to persistent storage before sending the PUBREC message.

ralight:
I'm not familiar with adversary schedulers, but the point stands - if something happens to stop the client before that call then the broker will not retry.

I'd be happy to receive a pull request for (optional) persistence support.

Issue #118, opened 2016-03-15
Support delaying PUBACKs in libmosquitto API


ralight:
migrated from Bugzilla [#452930](https://bugs.eclipse.org/bugs/show_bug.cgi?id=452930)
status NEW severity *enhancement* in component *Mosquitto client library* for *---*
Reported in version *unspecified* on platform *PC*
Assigned to: Roger Light

On 2014-11-23 17:37:39 -0500, Roger Light wrote:

> (imported from launchpad)
> 
> In the situation where a libmosquitto-based client subscribes to a topic with QoS=1, it would be useful to be able to delay the PUBACKs that it sends back.
> Instead of the current automatic behavior, the API would provide a callback to manually send a PUBACK for a specific message.
> The idea being that it confirms that the message has been correctly processed locally and is ready to be acknowledged to the broker.
> 
> The feature was already mentioned in a message from the mailing-list a year ago: https://lists.launchpad.net/mosquitto-users/msg00024.html
> 
> And I see that the MQTT.js implementation supports such a feature: https://github.com/adamvr/MQTT.js/pull/77

PR #1932, opened 2020-12-06
add mosquitto_delay_puback option

linuxbasic:
Signed-off-by: Linus Basig <linus@basig.me>
Signed-off-by: Fabrizio Lazzaretti <fabrizio@lazzaretti.me>

- [ ] ~~If you are contributing a bugfix, is your work based off the fixes branch?~~
  - `01-keepalive-pingreq.py` fails, but it also fails on master

-----

This PR adds a method `mosquitto_delay_puback`. After calling `mosquitto_delay_puback` the library will delay the PUBACK message required for QoS 1 message transfers until after the `on_message` callback returns. 

This change ensures the message does not get lost if the client crashes during the processing of the message. 

If the `mosquitto_delay_puback` is not called, the previous behavior (PUBACK before calling `on_message`) applies.

related issues: 
- https://github.com/eclipse/mosquitto/issues/118
- https://github.com/eclipse/mosquitto/issues/188

Our use-case is a messaging middleware that forwards the messages to other systems and we need to maintain the "At Least Once" delivery guarantee of QoS 1.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,4 @@ test/unit/out/
 
 www/cache/
 __pycache__
+.DS_Store
--- a/ChangeLog.txt
+++ b/ChangeLog.txt
@@ -1,3 +1,7 @@
+Client library:
+- Add `mosquitto_delay_puback()` for delaying the PUBACK message
+  until the on_message callback returns.
+
 2.0.0 - 2020-12-03
 ==================
 
--- a/include/mosquitto.h
+++ b/include/mosquitto.h
@@ -1665,6 +1665,22 @@ libmosq_EXPORT void mosquitto_user_data_set(struct mosquitto *mosq, void *obj);
  */
 libmosq_EXPORT void *mosquitto_userdata(struct mosquitto *mosq);
 
+/*
+ * Function: mosquitto_delay_puback
+ *
+ * Per Default, the library will acknowledge a QoS 1 message before it calls the 
+ * on_message callback. After <mosquitto_delay_puback> is called, the library
+ * delays the acknowledgment until after the on_message callback has returned.
+ *
+ * Parameters:
+ *  mosq - a valid mosquitto instance.
+ *
+ * Returns:
+ *	MOSQ_ERR_SUCCESS - on success.
+ * 	MOSQ_ERR_INVAL -   if the input parameters were invalid.
+ */
+libmosq_EXPORT int mosquitto_delay_puback(struct mosquitto *mosq);
+
 
 /* ======================================================================
  *
--- a/lib/cpp/mosquittopp.cpp
+++ b/lib/cpp/mosquittopp.cpp
@@ -352,6 +352,11 @@ void mosquittopp::user_data_set(void *userdata)
 	mosquitto_user_data_set(m_mosq, userdata);
 }
 
+int mosquittopp::delay_puback()
+{
+	mosquitto_delay_puback(m_mosq);
+}
+
 int mosquittopp::socks5_set(const char *host, int port, const char *username, const char *password)
 {
 	return mosquitto_socks5_set(m_mosq, host, port, username, password);
--- a/lib/cpp/mosquittopp.h
+++ b/lib/cpp/mosquittopp.h
@@ -107,6 +107,7 @@ class mosqpp_EXPORT mosquittopp {
 		int max_inflight_messages_set(unsigned int max_inflight_messages);
 		void message_retry_set(unsigned int message_retry);
 		void user_data_set(void *userdata);
+		int delay_puback();
 		int tls_set(const char *cafile, const char *capath=NULL, const char *certfile=NULL, const char *keyfile=NULL, int (*pw_callback)(char *buf, int size, int rwflag, void *userdata)=NULL);
 		int tls_opts_set(int cert_reqs, const char *tls_version=NULL, const char *ciphers=NULL);
 		int tls_insecure_set(bool value);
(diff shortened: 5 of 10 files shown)


karlp:
When did this change? I thought the library had always been pubacking after teh callback?

ralight:
@karlp I've checked, and it's been like this forever.

I'm mostly inclined to say that we should just change the behaviour (and for QoS 2) without an option. For people who have quick callbacks it will make no difference, and those with long callbacks there If would only potentially be a problem with non-compliant brokers.

If I did add it as an option, it would be with `mosquitto_int_option()` though, there is no need for a new function I don't think.

linuxbasic:
> When did this change? I thought the library had always been pubacking after teh callback?

It did not surprise me that the library acknowledges before the message is handed over to the application; most libraries do that. I think the reason for that is that the specification is not very explicit about this situation.

The [MQTT specifications](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_QoS_1:_At) says: 
> In the QoS 1 delivery protocol, the receiver MUST respond with a PUBACK packet containing the Packet Identifier from the incoming PUBLISH packet, having accepted ownership of the Application Message

`Figure 4.2 – QoS 1 protocol flow diagram, non-normative example` goes into more detail:
> The receiver does not need to complete delivery of the Application Message before sending the PUBACK. When its original sender receives the PUBACK packet, ownership of the Application Message is transferred to the receiver.

Personally, I'm also in favor of changing the default behavior. As @karlp's reaction shows the "acknowledging before processing" behavior is very counter-intuitive and the specification does not provide any good reason to use it.

Should I change the PR to send PUBACK and PUBCOMP after on_message was called?

karlp:
Hrm, I thought the paho python libs didn't ack until the callback returned, or something I was working with at least, I was explicitly demoing delays in the handlers.  I'd be perfectly fine with switching the default, but this will likely lead to beahviour changes as broken clients will now cause unacked packets to build up in varous places, whereas right now, a client that crashes doesn't matter.

linuxbasic:
We were using the [paho.mqtt.c](https://github.com/eclipse/paho.mqtt.c) library before and it has the same "issue": https://github.com/eclipse/paho.mqtt.c/blob/master/src/MQTTProtocolClient.c#L326-L332

The same for the [patho.mqtt.python](https://github.com/eclipse/paho.mqtt.python) does the acknowledging before the processing: https://github.com/eclipse/paho.mqtt.python/issues/348

There are libraries that implement a delayable acknowledgment like [paho.mqtt.java](https://github.com/eclipse/paho.mqtt.java/blob/master/org.eclipse.paho.mqttv5.client) (https://github.com/eclipse/paho.mqtt.java/blob/master/org.eclipse.paho.mqttv5.client/src/main/java/org/eclipse/paho/mqttv5/client/IMqttAsyncClient.java#L822-L833) and some that delay it per default like [mqtt.js](https://github.com/mqttjs/MQTT.js/) (https://github.com/mqttjs/MQTT.js/pull/697).

I think pending acknowledgments shouldn't be a big problem. They can already occur if there is an issue with the network layer or the client crashes before he reads the PUBLISH message from the network buffer. In any way, as soon as the connection's Keep Alive period runs out, these pending acknowledgments should be cleaned up.

ralight:
I think on balance the best thing to do is add it as an option otherwise if it does cause problems there is nothing users can do other than downgrade, if it does cause a problem. So please update this to use `mosquitto_int_option()` with say `MOSQ_OPT_DELAYED_ACK` as the option. Set it to off by default, and extend it to cover QoS 2 as well.

Then we go:

2.1: Option available, but off by default
2.2: Option enabled by default
3.0: Option removed?

linuxbasic:
I implemented the `MOSQ_OPT_DELAYED_ACK` option.

**Regarding QoS 2**:

(for the following thoughts I assumed `message__queue()` stores the message in memory and will be lost in the case of the process crashing, please correct me if I'm wrong)

With the current implementation of QoS 2, delaying the sending of the `PUBCOMP` until after the `on_message` exits would not have any effect because there is a mechanism in place that prevents `on_message` to be called more than once (see https://github.com/eclipse/mosquitto/blob/develop/lib/handle_pubrel.c#L101-L104).

As far as I see we have two options:
1. leave it like it is ("At Most Once" processing guarantee if the process crashes after the initial `PUBLISH` is received and `on_message` is called)
2. move the `on_message` callback call into the handler of the `PUBLISH` message (like the [MQTT specs](https://github.com/eclipse/mosquitto/blob/develop/lib/handle_pubrel.c#L101-L104) suggest) and implement the delayed ack logic there.
  - this would result in an "At Least Once" processing guarantee if we delay the acknowledgment (if the process crashes and loses the msg-id before `PUBREC` is sent).
  - this would result in a "0+" processing guarantee if we send `PUBREC` before calling `on_message` (0 in case the process crashes after `PUBREC` is sent and before `on_message` called and "1+" if the process crashes after `on_message` is called and at the same time there is a network issue that prevents the arrival of the `PUBREC`).

I personally am in favor of option 1 because QoS 2 sucks because the MQTT specification assumes that the sender and the receiver never crash or lose their state. Even if we go with option 2 we can't fulfill the promise of an "Exactly Once" processing guarantee in real-world conditions and option 1 is already as good as it gets if we ignore the option of a process crashing at the wrong moment.

linuxbasic:
@karlp @ralight What do you think?

Type1J:
Any news on when this might get reviewed/merged?

PR #2032, opened 2021-01-19
mosquitto_rr fixes

yamt:

-----

### DIFF
--- a/client/client_shared.c
+++ b/client/client_shared.c
@@ -498,7 +498,7 @@ int cfg_add_topic(struct mosq_config *cfg, int type, char *topic, const char *ar
 		}
 		cfg->topic = strdup(topic);
 	}else if(type == CLIENT_RESPONSE_TOPIC){
-		if(mosquitto_pub_topic_check(topic) == MOSQ_ERR_INVAL){
+		if(mosquitto_sub_topic_check(topic) == MOSQ_ERR_INVAL){
 			fprintf(stderr, "Error: Invalid response topic '%s', does it contain '+' or '#'?\n", topic);
 			return 1;
 		}
--- a/man/mosquitto_rr.1.xml
+++ b/man/mosquitto_rr.1.xml
@@ -632,7 +632,7 @@
 						<option>31</option>, or the more verbose
 						<option>mqttv5</option>, <option>mqttv311</option>, or
 						<option>mqttv31</option>.
-						Defaults to <option>311</option>.</para>
+						Defaults to <option>5</option>.</para>
 				</listitem>
 			</varlistentry>
 			<varlistentry>


yamt:
wrt "Have you successfully run make test with your changes locally?", i'm not sure.
i'm using cmake on macOS. "make test" says "No tests were found!!!".

client/client_shared.c
```diff
@@ -498,7 +498,7 @@ int cfg_add_topic(struct mosq_config *cfg, int type, char *topic, const char *ar
 		}
 		cfg->topic = strdup(topic);
 	}else if(type == CLIENT_RESPONSE_TOPIC){
-		if(mosquitto_pub_topic_check(topic) == MOSQ_ERR_INVAL){
+		if(mosquitto_sub_topic_check(topic) == MOSQ_ERR_INVAL){
```
ralight, line 501: This is an incorrect change. The response topic is telling the remote client where to publish a response, so it must not contain wildcards.

yamt: i don't understand. what's "the remote client"? from the POV of mosquitto_rr, it's a topic for subscribe.

ralight: Yes you are correct it is a subscribe topic for mosquitto_rr, but that doesn't mean it can be a subscription topic.

The way the request response feature works is that client A sends a PUBLISH as a request. That PUBLISH message has a response topic attached. The response topic tells client B, the client that will deal with the request, where to send the response. So it has to be an proper topic that can be published to, i.e. one that doesn't have wildcards in. That's why there is the check against it being a publish topic.

yamt: which "the request response feature" are you talking about?
i suspect i'm missing something obvious.
my motivation is documented in the commit message.
```
    It's necessary to talk with thingsboard, which
    requires a wildcard subscription. ("v1/devices/me/rpc/response/+")
    
    https://thingsboard.io/docs/reference/mqtt-api/#client-side-rpc
    (While not mentioned in the doc explicitly, it actually rejects
    non-wildcard topics like "v1/devices/me/rpc/response/1".)
```

ralight: Ah, I see. `mosquitto_rr` is a client that implements the `Request / Response` feature of MQTT v5, as described here: https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901252

I don't think it is suitable for use with thingsboard like that, you could try `mosquitto_sub`, but you wouldn't be able to reply. The best bet is to write a client that matches your needs I'm afraid.

yamt: ok. i was not aware of mqtt 5 request / reqsponse. thank you! let me learn a bit before replying.

Issue #2074, opened 2021-02-04
Mosquitto unable to start


jmichaud111:
Was creating items inside my OpenHab environment and all of a sudden everything went offline. Ive tried more then then I'd like to admit to try and resolve this but I ended up just removing all of eclipse and attemping to reinstall with (sudo apt-get purge --remove mosquitto*) but the error didn't change and I'm still dead in the water. Any insight or suggestions would be marvelous and much appreciated.

skynet@skynet:/etc/mosquitto$ sudo systemctl status mosquitto
● mosquitto.service - Mosquitto MQTT Broker
     Loaded: loaded (/lib/systemd/system/mosquitto.service; enabled; vendor preset: enabled)
    Drop-In: /etc/systemd/system/mosquitto.service.d
             └─override.conf
     Active: failed (Result: exit-code) since Thu 2021-02-04 15:36:03 EST; 6min ago
       Docs: man:mosquitto.conf(5)
             man:mosquitto(8)
   Main PID: 14891 (code=exited, status=1/FAILURE)

Feb 04 15:36:03 skynet systemd[1]: mosquitto.service: Scheduled restart job, restart counter is at 5.
Feb 04 15:36:03 skynet systemd[1]: Stopped Mosquitto MQTT Broker.
Feb 04 15:36:03 skynet systemd[1]: mosquitto.service: Start request repeated too quickly.
Feb 04 15:36:03 skynet systemd[1]: mosquitto.service: Failed with result 'exit-code'.
Feb 04 15:36:03 skynet systemd[1]: Failed to start Mosquitto MQTT Broker.

ralight:
Sorry you're having problems. Do you know what version of Mosquitto you have? `apt-cache policy mosquitto` should tell you.

You might get more information on what is going wrong by starting Mosquitto manually:

`sudo /usr/sbin/mosquitto -c /etc/mosquitto/mosquitto.conf`

exetico:
Same problem here.

Running the latest version in docker (2.0.7). My watchtower upgraded the container like it normally do, but everything was offline this morning.

The container is reporting "Address not available" - but I'm not sure if it's something related to IPv6 (My local setup are IPv4 only).

It's also reporting: "Starting in local only mode. Connections will only be possible from clients running on this machine"
And "Create a configuration file which defines a listener to allow remote access."

Therefore, something groundbreaking must have changed?

Esc4pe:
[...]

Issue #2040, opened 2021-01-21
docker eclipse-mosquitto:2.0 Error: Address not available


a-wing:
```bash
docker run eclipse-mosquitto:2.0
1611227315: mosquitto version 2.0.5 starting
1611227315: Config loaded from /mosquitto/config/mosquitto.conf.
1611227315: Starting in local only mode. Connections will only be possible from clients running on this machine.
1611227315: Create a configuration file which defines a listener to allow remote access.
1611227315: Opening ipv4 listen socket on port 1883.
1611227315: Opening ipv6 listen socket on port 1883.
1611227315: Error: Address not available
1611227315: mosquitto version 2.0.5 running
```

Error: Address not available

citrusvanilla:
geeze im glad i found this.  i tried to pin the 2.0.5 image as well in my docker-compose yaml and your message made me think it had to do with changing the tab.  i rolled back to 1.6.12 for the time being.  2.0.5 is giving me an error that i had suspected was from the mosquitto daemon possibly running in the background somehow though i could find no evidence supporting this.

ralight:
Mosquitto 2.0 binds to the loopback interface if you don't provide a `listener` configuration. That loopback interface probably isn't available to your docker container, so it fails. If you add `listener 1883` to your configuration it will start working happily again - but note that you also need to configure some form of authentication, or set `allow_anonymous true` if you don't want clients to have to authenticate.

Full details on moving to 2.0 are here: https://mosquitto.org/documentation/migrating-to-2-0/

Derkades:
Is there any way this can be the default for the docker image, for example an environment variable that is set by default in Dockerfile and detected by the mosquitto server? It is somewhat inconvenient to mount a config in the container just to get it to work, the previous version worked fine without any configuration.

Recently the :latest tag (which I admittedly shouldn't have been using) was switched from v1 to v2 and it broke my setup.

ralight:
@Derkades I'm sorry to hear it broke your setup.

The breaking changes have been advertised for the past two months, and the change of the `latest` tag was advertised a week before it happened. I appreciate that not everyone will want to / be able to follow all of the projects they are using for changes like this, but I hope you'll appreciate we have tried to give plenty of opportunity for people to discover that there are breaking changes, and to have chance to do something about them.

[...]

PR #2086, opened 2021-02-11
Update documentation

mosterdt:

-----

This pull request addresses some documentation changes regarding the migration to 2.0 and beyond.
There is also some discussion around this in #2074 and #2040. 

I will repeat here that while I agree that it's a great idea to only bind to the loopback interface by default, restricting the listener to localhost inside a docker container serves no benefit at all. It forces users to always define a mosquitto.conf. And even if mosquitto binds to every address within a docker container, it is still up to the user of the container to bind this port to their own network with e.g. `-p 1883:1883`.
Additionally, users need now also to explicitly allow anonymous users, or set up proper user authentication.

### DIFF
--- a/docker/2.0-openssl/README.md
+++ b/docker/2.0-openssl/README.md
@@ -33,6 +33,19 @@ to expose the ports that have been configured, for example:
 docker run -it -p 1883:1883 -p 8080:8080 -v <absolute-path-to-configuration-file>:/mosquitto/config/mosquitto.conf eclipse-mosquitto:<version>
 ```
 
+**Important**: The default configuration only listens on the loopback
+interface. This means that there is no way to access Mosquitto in the docker
+container without using a custom configuration containing at least a listener.
+You also need to make a decision to allow anonymous connections or to set up a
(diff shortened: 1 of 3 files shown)


mosterdt:
Removed the last commit about docker documentation as this was fixed in 12ff9d5e1491b805605d698c198b681db5c1dcbb

tobymurray:
I ended up here for exactly the reasons mentioned by @mosterdt. Docker Hub currently contains the phrasing:

> When running the image, the default configuration values are used. To use a custom configuration file, mount a local configuration file to /mosquitto/config/mosquitto.conf

I think this is much too passive, to the point where it misleads people. It seems like the default configuration should "work", so the default configuration yielding a roughly useless container is not a nice experience. 

I think the proposed changes (or similar) would be both an opportunity to reduce on-boarding friction for Docker users as well as give insight into the intended security configuration. Judging by #2074 it seems like a number of people are confused, so some prescriptive "this is what'd you'd need to do to get the equivalent of 1.* containers" would be welcome!

PR #2137, opened 2021-03-17
add client properties to plugins

dersimn:
## Before

Before it was not possible to obtain client information from within a message callback, see [example repository](https://github.com/dersimn/mosquitto-plugin-sandbox/tree/master):

```c
static int callback_message(int event, void *event_data, void *userdata) {
	struct mosquitto_evt_message *ed = (mosquitto_evt_message*)event_data;

	/*
		For some reason struct obtained with this 'Hack' is shifted.
		Maybe lib/mosquitto_internal.h -> stuct mosquitto is compiled with different preprocessor ifdefs?
	*/
	printf("by-struct id: %s\n", ed->client->id);                   // prints IP address
	printf("by-struct username: %s\n", ed->client->username);	// prints client id
	printf("by-struct password: %s\n", ed->client->password);	// prints username
	
	// workaround: Find base by tria & error:
	char **base = &ed->client->id;

	printf("by-base address: %s\n", base[0]);
	printf("by-base id: %s\n", base[1]);
	printf("by-base username: %s\n", base[2]);
	printf("by-base password: %s\n", base[3]);

	printf("payload: '%.*s'\n", ed->payloadlen, (char *)ed->payload);

	return MOSQ_ERR_SUCCESS;
}

// ...

mosquitto_callback_register(mosq_pid, MOSQ_EVT_MESSAGE, callback_message, NULL, NULL);
```

Output:
```
by-struct id: 192.168.161.1
by-struct username: someClient
by-struct password: someUser
by-base address: 192.168.161.1
by-base id: someClient
by-base username: someUser
by-base password: somePassword
payload: 'someMessage'
```

This is probably because we have no information about the struct `ed→client` (types are [mosquitto_evt_message](https://github.com/eclipse/mosquitto/blob/9f21a43eeeafdbb998409d075b986df23c18e649/include/mosquitto_broker.h#L141) → [mosquitto](https://github.com/eclipse/mosquitto/blob/9f21a43eeeafdbb998409d075b986df23c18e649/lib/mosquitto_internal.h#L209)) without including `mosquitto_internal.h` (which seems to be an internal header file, that is not supposed to be used from within a plugin). If you include this header file anyways using a [small hack](https://github.com/dersimn/mosquitto-plugin-sandbox/commit/b73ede6da94965e683f8e0d9abb5052018f90c0a) it leads to the shifting issue above, probably because we don't match all the `#ifdef` conditionals defined in [`struct mosquitto`](https://github.com/eclipse/mosquitto/blob/9f21a43eeeafdbb998409d075b986df23c18e649/lib/mosquitto_internal.h#L210).

## After
See [example repository](https://github.com/dersimn/mosquitto-plugin-sandbox/tree/mosquitto-mod):

```c
static int callback_message(int event, void *event_data, void *userdata) {
	struct mosquitto_evt_message *ed = (mosquitto_evt_message*)event_data;

	printf("client address: %s\n", ed->client_address);
	printf("client id: %s\n", ed->client_id);
	printf("client username: %s\n", ed->client_username);
	printf("payload: '%.*s'\n", ed->payloadlen, (char *)ed->payload);

	return MOSQ_ERR_SUCCESS;
}
```

This comes in handy for e.g. if you want to tag incoming messages with information about the publishing client, see example plugin ["add-user-properties"](https://github.com/dersimn/mosquitto/blob/efe1beb982dd8d2a1d5d48f3495a8d73835195d9/plugins/add-user-properties/add_properties.c#L49) which adds these information to outgoing messages as MQTTv5 user properties, resulting in messages like:
```javascript
Packet {
  cmd: 'publish',
  retain: false,
  qos: 0,
  dup: false,
  length: 143,
  topic: 'some/topic',
  payload: <Buffer 73 6f 6d 65 4d 65 73 73 61 67 65>,
  properties: {
    userProperties: [Object: null prototype] {
      foo: 'bar',
      foo_num: '42',
      '$timestamp': '1615989213367',
      '$client_id': 'someClientId',
      '$client_username': 'someUsername'
    }
  }
}
```

### DIFF
--- a/include/mosquitto_broker.h
+++ b/include/mosquitto_broker.h
@@ -151,6 +151,9 @@ struct mosquitto_evt_message {
 	uint8_t reason_code;
 	bool retain;
 	void *future2[4];
+	char *client_address;
+	char *client_id;
+	char *client_username;
 };
 
 
--- a/plugins/add-user-properties/CMakeLists.txt
+++ b/plugins/add-user-properties/CMakeLists.txt
@@ -1,0 +1,11 @@
+include_directories(${mosquitto_SOURCE_DIR} ${mosquitto_SOURCE_DIR}/include
+			${STDBOOL_H_PATH} ${STDINT_H_PATH})
+
+add_library(mosquitto_message_timestamp SHARED mosquitto_message_timestamp.c)
+set_target_properties(mosquitto_message_timestamp PROPERTIES
+	POSITION_INDEPENDENT_CODE 1
+)
+set_target_properties(mosquitto_message_timestamp PROPERTIES PREFIX "")
+
+# Don't install, these are example plugins only.
+#install(TARGETS mosquitto_message_timestamp RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}" LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")
--- a/plugins/add-user-properties/Makefile
+++ b/plugins/add-user-properties/Makefile
@@ -1,0 +1,27 @@
+include ../../config.mk
+
+.PHONY : all binary check clean reallyclean test install uninstall
+
+PLUGIN_NAME=add_properties
+
+all : binary
+
+binary : ${PLUGIN_NAME}.so
+
+${PLUGIN_NAME}.so : ${PLUGIN_NAME}.c
+	$(CROSS_COMPILE)$(CC) $(PLUGIN_CPPFLAGS) $(PLUGIN_CFLAGS) $(PLUGIN_LDFLAGS) -fPIC -shared $< -o $@
+
+reallyclean : clean
+clean:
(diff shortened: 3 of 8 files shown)


ralight:
Try this instead, using the `mosquitto_client_*()` functions defined in `mosquitto_plugin.h`.

```
static int callback_message(int event, void *event_data, void *userdata) {
	struct mosquitto_evt_message *ed = (mosquitto_evt_message*)event_data;

	printf("client address: %s\n", mosquitto_client_address(ed->client));
	printf("client id: %s\n", mosquitto_client_id(ed->client));
	printf("client username: %s\n", mosquitto_client_username(ed->client));
	printf("payload: '%.*s'\n", ed->payloadlen, (char *)ed->payload);

	return MOSQ_ERR_SUCCESS;
}
```

Thanks for spending your effort on this to no avail, I'm sorry you didn't spot those functions. The plugin interface needs better documentation.

PR #2178, opened 2021-04-14
Fix can't find openssl system variables on macOS

Richyeoh:
Fix can't find openssl system variables on macOS


Richyeoh:
All right, maybe it's not necessary,because I see cmake options in build options in CLion.

PR #2184, opened 2021-04-18
Fix building with Visual Studio 2008

bk138:
This older Microsoft compiler does not support mixing declarations and
code and misses some error defines. This commit enables building with
VS2008 by moving up some variable declarations and defining error
codes to their WinSock counterparts in case they're not defined.

Signed-off-by: Christian Beier <info@christianbeier.net>

- [ ] Have you successfully run `make test` with your changes locally? -> **Could not run as this is not available on Windows.**

-----


ralight:
Thanks Christian.

PR #2229, opened 2021-06-14
Support for bridge tcp user timeout option

abiliojr:
Please find the description of this change in #2230

### DIFF
--- a/man/mosquitto.conf.5.xml
+++ b/man/mosquitto.conf.5.xml
@@ -1645,10 +1645,10 @@ openssl dhparam -out dhparam.pem 2048</programlisting>
 						<replaceable>mqttv311</replaceable>.</para>
 				</listitem>
 			</varlistentry>
-
 			<varlistentry>
 				<term><option>bridge_tcp_keepalive</option> <replaceable>idle</replaceable> <replaceable>interval</replaceable> <replaceable>counter</replaceable></term>
 				<listitem>
+
                     <para>
                         Set TCP keepalive parameters for this bridge connection.
                         Use this option to allow you to set a long MQTT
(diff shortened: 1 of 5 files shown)


ralight:
This looks fine to me in principle, but I think it needs a bit more work.

* You disallow values == 0, but then seem to apply a value of zero if the option isn't set.
* I'd like to see more documentation around what this means, even if it is pointing to somewhere to read something else. Ideally some hints about what might be appropriate would be good.
* If this option isn't used, what is the OS default value? Without knowing that, or being able to find it out I don't know whether I need to reduce the value, or what it is reasonable to reduce it to.

I think the idea is good, it's just a bit opaque at the moment and so "nobody" would use it.

abiliojr:
>     * You disallow values == 0, but then seem to apply a value of zero if the option isn't set.
Changed to allow values of 0 (which translates to user default) or don't set the socket option by default.

>     * I'd like to see more documentation around what this means, even if it is pointing to somewhere to read something else. Ideally some hints about what might be appropriate would be good.
I added some extra information in the documentation/config example. Please let me know if it's enough.

>     * If this option isn't used, what is the OS default value? Without knowing that, or being able to find it out I don't know whether I need to reduce the value, or what it is reasonable to reduce it to.
Added a reference to the linux man page that specifies up to 20 minutes by default. This is the consequence of the default value of _tcp_retries2_ (15). Mentioned this relation in the mosquitto documentation.

I also added a [link to a page](https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/) explaining how this setting works, how to use it, and it's caveats.

@ralight , please let me know if these changes are enough

PR #2249, opened 2021-07-08
Swap bridge topic storage from array to linked-list

cdelston0:
These commits modifiy the storage of topics associated with a bridge from:
  a fixed size array allocated at config parse time
to: 
  a linked list which can be modified after config time without reallocating
  
These changes are required to enable upcoming work on 'transitive topics'.

### DIFF
--- a/lib/send_publish.c
+++ b/lib/send_publish.c
@@ -38,14 +38,13 @@ Contributors:
 #include "packet_mosq.h"
 #include "property_mosq.h"
 #include "send_mosq.h"
-
+#include "utlist.h"
 
 int send__publish(struct mosquitto *mosq, uint16_t mid, const char *topic, uint32_t payloadlen, const void *payload, uint8_t qos, bool retain, bool dup, uint32_t subscription_identifier, const mosquitto_property *store_props, uint32_t expiry_interval)
 {
 #ifdef WITH_BROKER
 #ifdef WITH_BRIDGE
 	size_t len;
-	int i;
 	struct mosquitto__bridge_topic *cur_topic;
 	bool match;
 	int rc;
@@ -64,8 +63,7 @@ int send__publish(struct mosquitto *mosq, uint16_t mid, const char *topic, uint3
 #ifdef WITH_BROKER
 #ifdef WITH_BRIDGE
 	if(mosq->bridge && mosq->bridge->topics && mosq->bridge->topic_remapping){
-		for(i=0; i<mosq->bridge->topic_count; i++){
-			cur_topic = &mosq->bridge->topics[i];
+		LL_FOREACH(mosq->bridge->topics, cur_topic){
 			if((cur_topic->direction == bd_both || cur_topic->direction == bd_out)
 					&& (cur_topic->remote_prefix || cur_topic->local_prefix)){
 				/* Topic mapping required on this topic if the message matches */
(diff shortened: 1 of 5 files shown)


src/mosquitto_broker_internal.h
```diff
@@ -507,6 +507,7 @@ struct mosquitto__bridge_topic{
 	char *remote_topic; /* topic prefixed with remote_prefix */
 	enum mosquitto__bridge_direction direction;
 	uint8_t qos;
+	struct mosquitto__bridge_topic *next;
```
ralight, line 510: Could you please put this earlier in the struct so there's no chance of any struct packing problems in the future?

src/bridge_topic.c
```diff
@@ -126,18 +162,21 @@ int bridge__add_topic(struct mosquitto__bridge *bridge, const char *topic, enum
...
-	if(topics == NULL){
+	cur_topic = mosquitto__malloc(sizeof(struct mosquitto__bridge_topic));
+	if(cur_topic == NULL){
 		log__printf(NULL, MOSQ_LOG_ERR, "Error: Out of memory.");
 		return MOSQ_ERR_NOMEM;
 	}
-	bridge->topics = topics;
+	cur_topic->next = NULL;
+
+	LL_APPEND(bridge->topics, cur_topic);
```
ralight, line 178: I think it would be better to append cur_topic to the list only once it is completely ready at the end of the function. With this as it is, we malloc cur_topic, then add it to the list, then if we get a strdup failure later on then we return with uninitialised memory in some of the struct members. Putting it at the end means if there is a failure we make sure we clean everything up and nothing bad goes on to the list.

lib/send_publish.c
```diff
@@ -64,8 +64,7 @@ int send__publish(struct mosquitto *mosq, uint16_t mid, const char *topic, uint3
 #ifdef WITH_BROKER
 #ifdef WITH_BRIDGE
 	if(mosq->bridge && mosq->bridge->topics && mosq->bridge->topic_remapping){
-		for(i=0; i<mosq->bridge->topic_count; i++){
```
ralight, line 67: Could you remove `int i` on line 48 now it's no longer in use please?

ralight (changes requested):
This mostly looks great, I've commented on a couple of very minor points that I would have just fixed myself if it wasn't for the other changes. Could you also please squash this into a single commit? It's not that complicated to have as a single one. You can force push to your branch and it will update the PR.

src/bridge_topic.c
cdelston0: Strong agree on only adding a fully initialised struct. Looking at it again, there are opportunities to leak memory which is assigned to cur_topic->topic and/or cur_topic->local_prefix if either of the bridge__create_prefix calls fail. Admittedly this is in the case of OOM, so not super dangerous, but incorrect nonetheless. Also more important when bridge__add_topic can be called during the brokers life and not just at config time.

lib/send_publish.c
cdelston0: Done

src/bridge_topic.c
cdelston0: I have modified bridge__add_topic to handle OOM cleanup in labelled error blocks at the end of the function. Heavy use of goto, which may not be to your taste. Let me know what you think.

ralight (approved):

Issue #1211, opened 2019-03-28
Unable to connect over webockets with TLS (wss) from Firefox


stepanmracek:
Hi,
I'm running Mosquitto v1.5.8 with enabled websockets, but I'm unable to connect from Firefox when the broker is running over TLS. Strange thing is that It works from Chrome/Chromium. I tried both MQTT.js as well as Paho javascript client libraries with no success.

Connecting always ends with error `Firefox can’t establish a connection to the server at wss://localhost:9001/mqtt.`

Environment:
- Mosquitto: 1.5.8
- libwebsockets: 3.1.0
- openssl: 1.1.1.b

Here is my broker configuration:
```
allow_anonymous false
password_file /home/stepo/tmp/mqtt_pwd

listener 9001
protocol websockets

cafile /home/stepo/tmp/mqtt-certs/ca.crt
certfile /home/stepo/tmp/mqtt-certs/server.crt
keyfile /home/stepo/tmp/mqtt-certs/server.key
```

(When I comment the cafile, certfile and keyfile entries and thus running on ws:/ everything in Firefox works)

Here you can find GIST for JS Paho client: https://gist.github.com/stepanmracek/9b6d4a6ee9ef9d4d6efe738c069a9577
I always outputs following error to the console: `AMQJS0007E Socket error:undefined.`

**HINT**
- I've also tried Mosquitto 1.4.11 with libwebsockets 1.6.3 and it *works*
- On another machine I have Mosquitto 1.4.15 but this time with libwebsockets 3.0.0 and there I'm facing the same issue -> *does not work from Firefox*.

Therefore I'm suspecting that this is some regression in libwebsockets that is unfortunately reproducible only in Firefox.

ralight:
Thanks for the hints, I'll take a look. I'd suggest using libwebsockets 2.4.x, that has proved to be a pretty reliable version.

timothy-godfrey:
Hi!
I can confirm similar behaviour: secure Websockets works between chrome/chromium and Mosquitto, but not between Firefox(v68) and Mosquitto, using Mosquitto v1.6.4 and libwebsockets-3.1. For the time being I will revert to libwebsockets-2.4

[...]

PR #2255, opened 2021-07-15
Disable HTTP2 when building libwebsockets for docker images.

tparker-usgs:
As reported in #1211, Firefox cannot connect over websocekts with TLS when using HTTP2. Without HTTP2 support in libwebsockets the connections fallback to HTTP and are able to complete.

This work around has been successfully used by multiple commenters and tests well. This pull requests implements the work around when building Docker images.

Signed-off-by: Tom Parker <tparker@usgs.gov>


ralight (approved):
This looks good to me, thank you very much. I'll see about getting it applied to the images straight away.

ralight:
https://github.com/docker-library/official-images/pull/10628

Issue #1052, opened 2018-11-22
Disconnect callback not called on TLS error


doragasu:
- Library version: 1.5.4
- Steps to reproduce:
  * Use asynchronous API and set a disconnect callback
  * Set up the test machine for a TLS error to occur (e.g. set the time to the epoch using `sudo date -s@0`).
  * Connect to the broker.
- Expected behavior: a TLS error occurs, and the disconnect callback is called.
- Current behavior: a TLS error occurs, but the disconnect callback is not called.

I fixed this problem with the following patch: https://github.com/doragasu/mosquitto/commit/95bde0bd10cb1bcc16e88942840553efa4afcee9
But I am not sure this is the best way to fix this problem.

doragasu:
This issue is still present in version 1.6.2 (but the patch in the comment above does not apply anymore).

karlp:
do you get a connect callback? I'm not sure I'd expect to get a disconnect callback if you never actually got connected in the first place (due to the tls error)

doragasu:
I get neither a connect nor a disconnect callback since version 1.5. I was getting one or the other on previous versions.

I understand that it might be strange receiving a disconnect cb if connection is not established, but when using asynchronous operation, the easiest way to know if connection did not progress is to get a callback called in these cases.

I could poll the status, but I would like to avoid polling for simplicity and performance reasons. if there is other way of doing this, I would be happy to read.

ralight:
Closed in #2317, thanks very much.

PR #2317, opened 2021-09-16
Call disconnect callback on TLS error.

doragasu:
When using the library asynchronously, TLS errors caused e.g. because
of the date/time not set correctly, cause the connection to be silently
dropped without the disconnection callback being invoked, as described
in issue #1052

This commit fixes the issue, returning the behavior to the one the
library had previous to version 1.5.4, if a TLS error occurs, the
disconnect callback will be invoked for the client to get notified of
the error.


ralight:
Thank you very much, this looks good to me.

PR #2401, opened 2021-12-09
mosquitto_property_copy_all: copy client_generated flag

diorcety:
The function mosquitto_property_copy_all doesn't copy the flag client_generated. If this function is used on the user side in order to construct mosquitto properties this will lead that only the first properties of the original properties will be send


ralight:
Thank you, I've made the change on the fixes branch now so am closing this.

PR #2451, opened 2022-02-07
tests: replace use of ssl.wrap_socket that throws warnings in Python …

ogayot:
…3.10
Hello,

The Python function `ssl.wrap_socket` [1], which is used at various places in the tests is deprecated starting Python 3.7 because it does not support hostname matching (which is considered insecure). In Python 3.10, the function now throws warnings at runtime, which makes Ubuntu & Debian autopkgtest fail.

The function `ssl.SSLContext.wrap_socket()` [2] comes in as the replacement and has support for SNI and hostname matching.

Replaced all uses of `ssl.wrap_socket()` by equivalent using `ssl.SSLContext.wrap_socket()` so that autopkgtest can now run with newer version of Python.

Thanks,

[1] https://docs.python.org/3/library/ssl.html#ssl.wrap_socket
[2] https://docs.python.org/3/library/ssl.html#ssl.SSLContext.wrap_socket

-----

-----


ralight:
Thank you very much, I wasn't aware of this deprecation.

PR #2536, opened 2022-05-16
Enable building on AIX

andreaskem:
This merge request allows for a simple build on AIX using GNU make with or without cmake. The mosquitto develop branch uses an OpenSSL version that is not yet packaged for AIX, so TLS and web sockets need to be disabled. On the master branch, WITH_TLS could still be used.

Only the persistence test suite is currently failing.

-----


ralight:
This looks great, and I'd be happy to add support for new platforms, but your comment about OpenSSL is a bit concerning. The develop branch works just fine with version OpenSSL 1.1.1, which is now 3.5 years old. Is it really the case that this isn't packaged for AIX? I've no experience with AIX, but this page suggests that 1.1.1 is available: https://www.ibm.com/support/pages/downloading-and-installing-or-upgrading-openssl-and-openssh

andreaskem:
You are 100% correct. This development machine I am on does still seem to  be using OpenSSL 1.0.2 for some reason. I foolishly thought that the missing symbols I encountered were related to a OpenSSL 3 migration in `mosquitto` because the `master` branch was building just fine and I was under the impression that this machine was kept up to date. I will have to talk to somebody about that and try to test with 1.1 but I do not assume that there would be an impact on this merge request.

andreaskem:
Apparently, and don't quote me on that, OpenSSL might not get updated to 1.1 automatically on AIX (yet). It is also possible that IBM is paying for extended support for OpenSSL 1.0.2, as is being offered on the OpenSSL website.

In any case, I was now able to test this merge request with OpenSSL 1.1 and could build including SSL/TLS support. I reworded the commit message accordingly and fixed some additional issues along the way (a warning in `src/context.c` and two `install` failures).

Unfortunately, `set-version.sh` is using the non-POSIX option `-i` and AIX sed does not implement that. I worked around that locally by modifying my `PATH` to point to GNU sed instead.

plugins/persist-sqlite/Makefile
```diff
@@ -23,6 +23,12 @@ else
 ALL_DEPS:=
 endif
 
+ifeq ($(WITH_CJSON),yes)
+ALL_DEPS:=$(ALL_DEPS)
+else
+ALL_DEPS:=
+endif
+
```
karlp, line 31: these two cjson patches seem unrelated to building on aix, are they just fixes for "building without json support" ?

src/CMakeLists.txt
```diff
@@ -261,6 +261,7 @@ if(UNIX)
 		set_target_properties(mosquitto PROPERTIES
 			LINK_FLAGS "-Wl,-exported_symbols_list -Wl,${mosquitto_SOURCE_DIR}/src/linker-macosx.syms"
 		)
+	elseif (${CMAKE_SYSTEM_NAME} MATCHES "AIX")
```
karlp, line 264: shouldn't you be adding your linker-aix.syms here too?  (Or make it emit a warning that cmake isn't supported on AIX?)

src/net.c
```diff
@@ -651,7 +653,7 @@ int net__tls_load_verify(struct mosquitto__listener *listener)
 }
 
 
-#ifndef WIN32
+#if !defined(WIN32) && !defined(_AIX)
```
karlp, line 656: This means you should probably update https://github.com/eclipse/mosquitto/blob/master/man/mosquitto.conf.5.xml#L1096 to include AIX there too.

src/CMakeLists.txt
andreaskem: `cmake` seems to generate an export file and adds some AIX export flags on its own. Using `make VERBOSE=1` I noticed:
```
-Wl,-bE:CMakeFiles/libmosquitto.dir/objects.exp
```
I have no idea how `cmake` handles this stuff but the build artifacts seem to work just the same.

src/net.c
andreaskem: Sure.

src/CMakeLists.txt
karlp: cool :) I'm not a cmake person either, just looked wrong here

andreaskem:
Sorry for bothering you, but is there a blocker for this pull request or do I just have to be patient?

ralight:
I've updated to the latest develop branch and fixed the conflicts. Note that develop now has cjson as a requirement so it may still not work

Issue #2594, opened 2022-07-19
[BUG] server disconnects the TCP connection while TLS connection is not complete, client will dead loop


CastleOnTheHill:
https://github.com/eclipse/mosquitto/blob/ff97fbfe94b304b86167b15a68d4ae1d4502642c/lib/net_mosq.c#L576

mosquitto version: 1.6.7
openssl version: 1.1.1k
platform: linux 4.14

This problem still exists in the latest version。

Hi, I'm working with mosquitto and i found: When the TCP connection is complete, but the TLS connection is not complete. If the server disconnects the TCP connection,  In this case, SSL_get_error returns the error code SSL_ERROR_SYSCALL.
In the current implementation, if the error code is SSL_ERROR_SYSCALL, the mosquitto_loop function returns success and continues to invoke SSL_connect in the next mosquitto_loop,

Because the TCP connection is disconnected, SSL_connect immediately returns the failure error code SSL_ERROR_SYSCALL and invoke the next mosquitto_loop again. As a result, the program enters a dead loop.
This dead loop causes high CPU usage. In fact, the mosquitto_loop does not return failure until the mqtt keepalive time expires.

According to https://www.openssl.org/docs/man1.1.1/man3/SSL_get_error.html
> SSL_ERROR_SYSCALL
Some non-recoverable, fatal I/O error occurred. The OpenSSL error queue may contain more information on the error. For socket I/O on Unix systems, consult errno for details. If this error occurs then no further I/O operations should be performed on the connection and SSL_shutdown() must not be called.

`SSL_ERROR_SYSCALL` is fatal error，when `SSL_get_error` return `SSL_ERROR_SYSCALL`, we should return an error to the user。

CastleOnTheHill:
I merge [Backport SSL connect fixes.](https://github.com/eclipse/mosquitto/commit/e979a46c048a8c60c53614548c4a98dfd4992cf4) in my local workspace。program 
passed the above situation successfully, and the code is indeed a lot cleaner。

ralight:
Great, thank you very much for confirming.

PR #2595, opened 2022-07-19
Fix SSL_connect misuse

CastleOnTheHill:
Fix issue https://github.com/eclipse/mosquitto/issues/2594#issue-1309240883
SSL_ERROR_SYSCALL is fatal error, should not return success to user


ralight:
Thank you for your PR, I decided to backport the same fix from the `develop` branch instead, I hope that is ok.

CastleOnTheHill:
It is ok，your fix [Backport SSL connect fixes.](https://github.com/eclipse/mosquitto/commit/e979a46c048a8c60c53614548c4a98dfd4992cf4) is better。

PR #2619, opened 2022-08-19
support colon in username in pwfile via double colon

stefanschuermans:
There are smart IoT devices that use the string represeantation of the MAC
address as MQTT user name, i.e., a user name containing colon characters.
In basic setups of Mosquitto with authentication based on pwfile, user names
could not contain colon characters, because the colon is used as a delimiter.

This commit adds supports for colon characters in user names in pwfile, by
interpreting double colons in the user name in pwfile as a single colon in
the actual user name.

Signed-off-by: Stefan Schuermans <stefan@schuermans.info>

- [ ] ~If you are contributing a bugfix, is your work based off the fixes branch?~

-----


rswindell:
Isn't it true that the pwfile syntax only allows for a single ':' delimiter character on each line? The password is a base64-encoded hash, so there should be no colons in the password field. So this would mean just checking for the *last* occurring colon (e.g. using `strrchr()`) would give you the delimiter between the username and password fields. With that agreement, this change could be made much simpler (no strtok usage or re-implementation needed, no new dynamic memory allocation).

stefanschuermans:
I have to admit that I did not find the specification of the pwfile syntax. I think the current implementation would take the part between the first colon and the end of the line or the second colon as the password field. I did not want to break the potential option to append additional fields in the future. However, I have no real idea what those could be used for.
Your suggestion of using the last colon in the ling to split username/password would solve my issue and it would be way less code. I can change the implementation to use this approach.

PR #2683, opened 2022-11-09
Fix memory leak in mosquitto_tls_opts_set()

guillaumepellegrino:
When calling mosquitto_tls_opts_set() multiple time in a row on the same mosquitto context (Example: trying to reconnect to mqtt broker), some of the tls options are not free, resulting in a memory leak.

The TLS options (tls_version, tls_ciphers) must be free before being set.


rswindell:
Sounds like additional tests (using valgrind?) need to be added as well.

PR #2689, opened 2022-11-21
fix example JSON

zivillian:
The invalid JSON is answered with:
```json
{
  "responses": [
    {
      "command":"Unknown command",
      "error":"Payload not valid JSON"
    }
  ]
}
```


ralight:
Thank you.

PR #2698, opened 2022-12-02
Use thread-safe/reentrant version of gmtime(): gmtime_r()

rswindell:
gmtime() is not thread-safe.

Ideally, this code would check the return value of gmtime_r() for NULL (indication of failure), but since this wasn't being done before, I didn't want to overload this commmit with an extra "feature".

Signed-off-by: Rob Swindell <rob@synchro.net>

No, the "fixes" branch currently doesn't actually build for me (logging.c:294:96: error: expected ‘)’ before ‘PRIu64’) - a different issue.
-----

### DIFF
--- a/plugins/message-timestamp/mosquitto_message_timestamp.c
+++ b/plugins/message-timestamp/mosquitto_message_timestamp.c
@@ -44,15 +44,15 @@ static int callback_message(int event, void *event_data, void *userdata)
 {
 	struct mosquitto_evt_message *ed = event_data;
 	struct timespec ts;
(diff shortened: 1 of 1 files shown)


ralight:
Thanks for the suggestion, but I'd prefer to stick with plain `gmtime()` because it is more portable - `gmtime_r()` is not on Windows, and `gmtime_s()` annoying has its parameters swapped - this could be worked around, but at the expense of confusing the example.

This code is all single threaded, so I think it is fine as it is.

PR #2738, opened 2023-01-25
Add yaml support to dynamic security plugin

axos88:

-----

Adds yaml serialization / deserialization to the dynamic-security-plugin so that the config file can be more easily be understood and modified by a human in case one does not actually want to use the dynamic part of the plugin (or not exclusively).

This is a PoC, would appreciate some preliminary feedback.

A lot of debug/trace code is still there and it needs a lot of cleanup, but it's enough to get some idea on what I'm trying to do.

Some notes / open questions / TODOs:

  - [x] Makefile changes need to be done - No guards yet for the availability of the YAML library.
  - [x] The plugin auto-guesses the format based on the file extension, although this could be an (optional) plugin parameter as well. I don't see why having to suffix yaml files with .yaml or .yml would pose an issue, but it might, so it might be worthwile to add an ovveride parameter.
  - [ ] One could argue to remove the json loaders, since the yaml parser can parse the json files, however this would mean that in order to use the plugin, one would need the yaml library.
  - [ ] The yaml parser can only parse the data in the order they appear in the file. Some changes needed to be made in case the client and group lists appear earlier than the roles in the file, as otherwise the system would err out with role not found errors.
  - [x] Error handling need to be cleaned up, and corrected, currently all yaml functions return 1 for success, 0 for error. The parse errors themselves are written to the log more-or-less correcly.
  - [ ] No testing yet, manually checked that a few json files are read (by the yaml deserializer!) and written back to yaml correctly. Some mosquitto_ctrl commands have been issued.
the initialization of the plugin is currently written out as json, and I don't intend to change that. The yaml parser will parse that json file correctly anyways.
  - [x] Probably a lot of memory handling issues are still present, will need to verify that no memory is leaked. Gah, I hate C.

My plans in the future (not part of this effort):

  - Answer the question on what happens if another auth plugin accepts a user that is not registered as a client in the dynamic security plugin. (currently have an hmac authentication plugin where the broker doesn't need to know about all its clients beforehand, so no configuration is needed on the broker to have a new client connect - same case with signed client certificates)
  - Add the ability to encode groups in usernames (any user connectiong with g1-foo will be part of the group g1 if it exists.). This will allow other plugins to authorize users, and still have the dynamic security plugin work for them. (Or another smart way to handle these)


axos88:
This is now ready for the first round of review.

ralight:
After a great deal of consideration, I'm afraid I'm going to close this request without merging it. I really find it hard to reject changes when people have put considerable effort in, however I don't believe this is a good fit as we have discussed, particularly around the future maintainability. Other people in the community I have spoken to have expressed similar concerns.

Thank you again for the effort that you have expended on this, I hope at least that you found doing so interesting or enjoyable, even if this isn't the outcome you wanted.

PR #2740, opened 2023-01-31
Clientid prefixes should support username wildcard

axos88:

-----

Adds support for clientid prefixes such as "%u-" to support prefixing each clientid with the connecting username


ralight:
This is an interesting idea - I like the concept of tying the username and clientid together, as seen on other config options, so that the end client has less control of what they can set as a clientid.

What use case do you see for this change? For some context for you, the client id prefix option exists to support MQTT v3.0, when the protocol did not have username/password support. Given that there are very few people who are aware of the ins and outs of v3.0, I was considering marking this option as deprecated and for removal in a future version. If there is still a good reason to keep it I'm all ears. I've just checked and v3.1 was about 13 years ago now, so any client library that might rely on this feature for crude authentication and doesn't support v3.1 must no longer be maintained, so it isn't something we would otherwise need to keep around.

axos88:
I'm a bit concerned that an authorized, but malicious user could disrupt the service for others if they know the client IDs in use.

Suppose malice has a valid user/pass to the system. If they learn bobs client I'd, they could kick him off the server by initiating an matt connection with their own user/pass, but bobs client id. 
Also could be used to leak information about what bob is up to: if they keep the connection open they will know when bob is attempting to use the server, because by doing so bob will kick malice's connection, this malice now knows that Bob is doing something related to the mqtt server.

Both issues would be solved by refusing any connection where the clientid doesn't start with, end with, or simply contain the username.

Issue #2722, opened 2023-01-05
TLS_version behavior does not match documentation


KramNamez:
The fix from https://github.com/eclipse/mosquitto/issues/2110 does not match the documented behavior that only TLSv1.2 and v1.3 are enabled if no `tls_version` is specified. Instead, v1.1 is also enabled.

ralight:
Thank you, this is fixed in the `fixes` branch and will be part of 2.0.16.

PR #2741, opened 2023-02-01
Add SSL_OP Flag For TLS Version Selection

Hyeongon-Kim:

-----

fix a bug that performs a handshake with tls1.3 even when the tls-version option is set to tlsv1.2 (or tlsv1.1)

### DIFF
--- a/.github/workflows/delete-old-workflow-runs.yml
+++ b/.github/workflows/delete-old-workflow-runs.yml
@@ -1,0 +1,18 @@
+name: Delete old workflow runs
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 1 * *'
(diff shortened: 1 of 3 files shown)


Hyeongon-Kim:
i have checked the old issues,
but i think we should be able to select tls-version clearly for *special situations.
* situations such as implementing and using specific crypto-algorithms based on tls-engine

Hyeongon-Kim:
#2722 i added code for this issue

ralight:
Thank you for this. The issue at #2722 is valid, and I have just added a change for that. What you are proposing here though is not correct in my opinion. The tls version options set a minimum version, whereas your changes set an explicit version.

Is there a reason why setting a minimum version is problematic for engine support?

PR #2742, opened 2023-02-02
fixed a bug where client could not load tls-engine

Hyeongon-Kim:
when loading tls-engine from the client,
if client_opts_set is called before client_connect, an error 'setting TLS engine' occurs. so i modified to call the net__init_tls function in the mosquitto_lib_init function.

Signed-off-by: Hyeongon Kim <khyeongon@gmail.com>

-----

when loading tls-engine from the client,
if client_opts_set is called before client_connect, an error 'setting TLS engine' occurs.
so i modified to call the net__init_tls function in the mosquitto_lib_init function.

-----
before : 

client engine before

after : 

client engine after

### DIFF
--- a/lib/mosquitto.c
+++ b/lib/mosquitto.c
@@ -80,6 +80,10 @@ int mosquitto_lib_init(void)
 		if (rc != MOSQ_ERR_SUCCESS) {
 			return rc;
 		}
+		
+		#ifdef WITH_TLS
+			net__init_tls();
+		#endif
 	}
 
 	init_refcount++;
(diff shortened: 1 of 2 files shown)


ralight:
Thank you for looking at this, the TLS engine code isn't currently covered by tests and isn't something I've spent a huge amount of time on.

The reason that the TLS initialisation happens where it does is to avoid always that initialisation if TLS is not in use. I do not want to move `net__init_tls()` back to the general lib init function. Could we instead add a call to `net__init_tls()` in `mosquitto_string_option()` for the `MOSQ_OPT_TLS_ENGINE`? I think that should satisfy both of us.

Do you have any experience of how to port engine related code to openssl 3? I haven't spent any time looking into that side of things and would definitely appreciate some insight.

PR #2743, opened 2023-02-02
Mosquitto ctrl allow dynsec client create without password

axos88:

-----


apps/mosquitto_ctrl/dynsec_client.c
```diff
@@ -56,6 +56,13 @@ int dynsec_client__create(int argc, char *argv[], cJSON *j_command)
 			}
 			password = argv[i+1];
 			i++;
+			request_password = false;
+		} else if (!strcmp(argv[i], "--no-password")) {
+			if (password != NULL) {
+				fprintf(stderr, "Error: -p and --no-password are mutually exclusive.\n");
```
ralight, line 62: You've covered the error case of `-p password --no-password`, but not `--no-password -p password` - could you add that check as well please (checking on `request_password` should do it)

ralight (changes requested):
I think this is a good change, thank you. I've made a specific comment on one of the changes. In addition to that, could you please add the new option to the help in `dynsec__print_usage()`?

If you could also add some entries to `test/apps/ctrl/ctrl-dynsec.py` in the `# Create clients` - the valid case, and then the two conflicting cases with the `-p` option as well that would be great.

Thanks again, this looks useful to me.

apps/mosquitto_ctrl/dynsec_client.c
axos88: Good catch. Will do in a few days.

PR #2751, opened 2023-02-13
Link with correct variant of websockets library

diizzyy:
Upstream refers to static library as websockets and shared as websockets_shared

Reference:
https://github.com/warmcat/libwebsockets/blob/main/lib/CMakeLists.txt#L279


diizzyy:
Feel free to commit this just fix the bug, I dont need to be credited

ralight:
Thank you for the suggestion. LWS used to be licensed under LGPL with a static linking exception. In order to satisfy licensing requirements when directly distributing mosquitto including websockets support, the library was static linked. This situation existed only really existed on Windows, and on Windows the resulting static library is renamed as libwebsockets_static, hence why this exists as it does.

As LWS decided to relicense to MIT so this problem doesn't exactly exist any more, but anyway I think the option can stay as it is for 2.0.x.

PR #2756, opened 2023-02-21
allow example plugins to be installed if variable INSTALL_XXX_PLUGIN …

axos88:

-----

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -139,6 +139,10 @@ WITH_SQLITE=yes
 # currently only suitable for use with oss-fuzz.
 WITH_FUZZING=no
 
+# To also install example plugins,
+# uncomment the following line and replace XXX with the SCREAMING_SNAKE_CASE of the example plugin.
+# INSTALL_XXX_PLUGIN=yes
+
 # =============================================================================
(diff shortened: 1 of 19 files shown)


ralight:
I think having the option of installing the example plugins is useful, however in the interests of keeping complexity down I think it would be best to have `WITH_EXAMPLE_PLUGINS` - so an all or nothing option that enables/disables the examples in `plugins/Makefile` and `plugins/CMakeLists.txt`. Disabled by default, and if enabled, they are also installed.

What do you think?

If you are have made this PR because you are interested in using one or more of the examples in a real use case, then I don't think selective installation of examples as you propose is the right choice. In that case, looking at promoting the examples to be "real" plugins, including tests, would be a better approach.

axos88:
My intention for the plugin was indeed to have the ability to create docker images with a few of the example plugins. I'm kinda stuck in the middle of a big refactor, but I'll take a look at which ones they actually were. 

It is possible that it might be worth considering promoting them as real plugins. The all or nothing build option might be a good solution as well. Then one could decide which so files to keep in the final image if the intent is to keep image size down.

Issue #2760, opened 2023-02-25
After the device system sleeps and wakes up again, the PING request is not sent


liangliang-zhll:
hi:
I have a problem here, please help me guide, thank you!
I am using libmosquito to develop mqtt client applications;using mosquitto_connecte () to connect the mqtt broker, and set keepalive time is 3 minutes. When a PINGREQ request is sent and the PINGRESP response message is received, I trigger to let my device system sleep for 2 minutes, and then wakes up the device system. After waiting for 1 minute, the PINGREQ request is not sent;

I found mosquito_ loop_ misc()--> mosquitto__ check_ keepalive() --> mosquitto_ time() --> clock_ gettime(CLOCK_MONOTONIC, &tp); The current time obtained is the system running time. When the system sleeps, it will not be counted. So when the system wakes up, the two minutes of the system sleep will not be counted, so waiting for one minute will not send a PINGREQ request;

Excuse me: 1) Does mosquito not support the keepalive function of the system sleep scenario? 2) Is there any good solution to this problem?
(Can we use clock_gettime (CLOCK_BOOTTIME,&tp) to solve the above problem?)

Please help to answer it, thank you!

Daedaluz:
Interesting edge-case...

While i think using `CLOCK_BOOTTIME` when possible is generally a good idea and would probably solve this specific test in your case, I would not rely on the connection being preserved over sleeps.

There are other layers of timeout implementations that also needs to be considered (like TCP and TLS for example).

If this is a system where you are using sleep periodically to preserve battery or something similar,
I would instead of trying to keep a specific connection alive over sleeps, just make use of persistent clients and do new connections when required and disconnect over sleep. 

The only thing this change would do is _maybe_ let the client use the same connection after a sleep, provided that the server hasn't already cut the connection while the client were sleeping.

Also: [this](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=d6ed449afdb38f89a7b38ec50e367559e1b8f71f) suggests that the issue is gone in linux 4.17 and later.

1) I'm not sure if this is about mosquitto supporting system sleep or not, given that `CLOCK_MONOTONIC` _should_ count sleep time too, and that mosquitto should do reconnects (if configured to do so) when the connection is lost (for whatever reason) + mosquitto supporting QoS 1, 2 should prevent any messages having qos > 0 beeing lost during temporary uplink issues.
[...]

PR #2762, opened 2023-03-02
use CLOCK_BOOTTIME when available.

Daedaluz:
This is just a minor fix for #2760


liangliang-zhll:
For this modification, About when can be merged?

ralight:
Thank you, this will be in 2.0.16 in the next few weeks.

PR #2768, opened 2023-03-08
 fixes build fail on darwin arm64

cicorias:
made cmake same as other plugin
examples and now it compiles
clean with cmake


cicorias:
my email on Eclipse foundation  matches the signature on the commit. Was just changed this evening, so perhaps it's pending. 
image

ralight:
That's lovely, thank you. I've verified the ECA manually, it's all good.

PR #719, opened 2018-03-12
Add PKCS11 and file URI support to Mosquitto

vareddy:
These changes allows certs and CAs to be in the form of file URIs and private keys to be either file URIs or PKCS11 compliant URIs.The private keys can either be stored on the filesystem or in a PKCS11 compliant Trusted Platform Module/Hardware Security Module (TPM/HSM). This provides more security to the private keys as they cannot be directly copied off of the filesystem.
Changes:
1. Added new API mosquitto_tls_set_uri to accept cert and CA as file URI (RFC8089) and private key as PKCS11 compliant (RFC7512) URI.
2. The above function also accepts paths to the libp11 library and the PKCS11 provider library, which is required to access keys stored on a TPM/HSM
3. Modified internal tls_set function to parse URI
4. Modified internal ssl connect function to initialize and call appropriate OpenSSL engine functions to load private key from the HSM/TPM
5. Added options for mosquitto_sub and mosquitto_pub to accept URIs and lib paths
6. Added unit tests for the new mosquitto_tls_set_uri function

The above changes were tested using SoftHSM2 as the keystore.

Signed-off-by: Varun Reddy <vareddy@amazon.com>

PR #2827, opened 2023-06-03
Fix engine keyfile support.

bdesplanq:
This small patch fixes two outstanding issues with OpenSSL engine support in the MQTT client:
1. When the tls_keyform is "engine" then the "keyfile" representing the private key is not neccessarily a real file but can be any handle known to the engine (URL, number, ...).
When the keyfile argument is validated in the function `mosquitto_tls_set()` then it should not be tried to open this file unless tls_keyform is "pem".
Because the tls_keyform must be known before calling `mosquitto_tls_set()`, the parsing of this tls_keyform config setting needs to be done before parsing the capath/cafile settings. 
2. Dynamically loadable OpenSSL engines require the `OPENSSL_INIT_ENGINE_DYNAMIC` option to be set or the `ENGINE_by_ID()` call will fail to find the engine in `mosquitto_string_option()`. As indicated in the OpenSSL documentation this option is not set by default.

This fix can also be seen as a lightweight and generic alternative to #719 which is fully oriented towards the pkcs11 engine while there are many other engines out there.
The `mosquitto.conf.5` man page describing the tls_keyfile conf setting has also been updated because the current description is only correct when tls_keyform is "pem".

-----

### DIFF
--- a/apps/mosquitto_ctrl/options.c
+++ b/apps/mosquitto_ctrl/options.c
@@ -593,6 +593,11 @@ int client_opts_set(struct mosquitto *mosq, struct mosq_config *cfg)
 		return 1;
 	}
 #ifdef WITH_TLS
+	if(cfg->keyform && mosquitto_string_option(mosq, MOSQ_OPT_TLS_KEYFORM, cfg->keyform)){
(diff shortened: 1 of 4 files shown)


ralight:
Looks good, thank you.

PR #2841, opened 2023-07-05
Fix define to properly parse bridge tcp user timeout configuration

SergeyVasiliev:
Parsing "bridge_tcp_user_timeout" configuration value is done under 'TCP_USER_TIMEOUT' define but is should be 'WITH_TCP_USER_TIMEOUT' instead as the latter used everywhere.


ralight:
Thank you.

PR #2844, opened 2023-07-06
Enable log filtering

a-turko:
Enable users to decide for what kinds of log messages their log callbacks will be called. This filtering happens in the library in
order to avoid the malloc call in log__printf (logging_mosq.c) for messages that would be ignored anyway.

By default, no messages are filtered out. The set of allowed log types can be configured by mosquitto_log_levels_set.

Also, this pull request adds a unit test of this feature.

Note:
Two tests failed locally: 06-bridge-clean-session-csF-lcsT.py and c/03-publish-c2b-qos2-maximum-qos-1.test, but those were already failing before these changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,6 +109,7 @@ test/unit/broker/bridge_topic_test
 test/unit/coverage.info
 test/unit/broker/keepalive_test
 test/unit/lib/lib_test
+test/unit/lib/logs_test
 test/unit/broker/persist_read_test
 test/unit/broker/persist_write_test
 test/unit/broker/subs_test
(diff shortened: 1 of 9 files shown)


a-turko:
Is there something wrong with this PR? Should I have included the requirement checklist in the message?
Is there anything else I could do to get the approval for the workflows?

a-turko:
@ralight, what do you think about this change?

> Is there something wrong with this PR? Should I have included the requirement checklist in the message? Is there anything else I could do to get the approval for the workflows?

bgianfo:
I just was reading the code and realized this was an issue. This feature would be awesome if we could get it merged into the next release. Avoiding these extra allocations would be a great improvement.

a-turko:
> I just was reading the code and realized this was an issue. This feature would be awesome if we could get it merged into the next release. Avoiding these extra allocations would be a great improvement.

Thank you. However, so far this PR has not been getting any attention from the maintainers, so for now I'm going to hold off with making changes.

Issue #2546, opened 2022-05-23
persistent_client_expiration only works within one run of mosquitto


JSchy65:
mosquitto broker version 2.0.14 (Ubuntu 20.04)  and 1.6.15. (CentOS 8)

The promise, that clients are removed if the persistent_client_expiration time is exceeded, is only kept for clients that are disconnected during the current run of the mosquitto broker.

The consequence of this is, that if the mosquitto brokers uptime is less than the persistent_client_expiration time, clients that are disconnected during that run are NEVER removed.

Imagine that someone has configured a persistent_client_expiration e.g. for one week. If, for some reasons, the mosquitto is restarted before the week is over, the clients disconnected during the first days of that week will NEVER be removed and their related messages will be stored FOREVER.

In my opinion, this makes the usage of that option useless (if long periods of time are configured for that parameter). Or it gives the user a false sense of security that his broker will never be flooded with messages no one is waiting for anymore.

In my opinion, the solution is to consider ALL disconnected clients for removement if the persistent_client_expiration is exceeded.

Here is a short excerpt from the log file to show the situation.
See the attached log file for more details.

[mosquitto-v2.0.14.log.zip](https://github.com/eclipse/mosquitto/files/8756165/mosquitto-v2.0.14.log.zip)

Configuration:

persistent_client_expiration 1h
persistence true
persistence_location /var/lib/mosquitto/

Log excerpt:

--------------------------------
1st run:

15:19:43
broker/version mosquitto version 2.0.14
broker/uptime  88 seconds
clients/total   1
messages/count 53

15:21:18
create 25 clients with clienIds 100..124

15:23:21
disconnect 25 clients 100..124

15:23:39
send 500 messages

15:24:01
uptime         253 seconds
clients/total   26
messages/count 553

--------------------------------
2nd run:

15:26:47
restart mosquitto

15:27:19
create 50 clients with ClientIds 200..249

15:27:45
disconnect 50 clients 200..249

15:28:47
send 300 messages

15:28:59
broker/uptime  132 seconds
clients/total   76 (+ 50 clients since 1st run)
messages/count 853 (+300 messages since 1st run)

16:27:46
removed clients 200..249 "Expiring client due to timeout" <- as expected 1h after disconnect

16:39:09
uptime         4334 seconds
clients/total   26 !!! (clients from the 1st run still exist)
messages/count 852 !!! (messages for the clients from 1st run still exist)

->
Only the clients 200..249 disconnected during the 2nd run were removed.
[...]

PR #2859, opened 2023-07-22
Remove session expiration timer upon setting up bridge connection (fix #2634)

### LINKED ISSUE #2634: 2.0.15 regression: bridge is not created, with "Expiring client local.xxx due to timeout"
After upgrading to 2.0.15, we observed that in about 50% of cases broker bridges are not created on start-up. The log shows the message
```
Expiring client local.<bridge-name> due to timeout.
```
immediately on start-up (i.e. not after an actual conceivable timeout).

This was first reported in https://github.com/eclipse/mosquitto/issues/1334#issuecomment-1219606937 and subsequent comments in that thread (thanks @everactivetim, @vomov), though it's almost certainly a separate issue from #1334. It appears to have been an unintentional side effect of https://github.com/eclipse/mosquitto/commit/c99502a2567bf9147b3d811cb17e38ad51f08660. Indeed, the issue does not seem to occur in 2.0.14.

In our specific case this is on Linux on ARM.

swiergot:

-----

This fixes #2634 by removing a possibly existing session expiration timer when setting up a bridge connection.

The bug is a side effect of the fix to #2546. Normally session expiration timers are not created for bridge connections. Starting with 2.0.15 the timers are created at startup for all clients found in the persistent database which also has records for bridge connections.

I have not worked with mosquitto code before, but I can see a few potential solutions:

- Don't create session expiration timers for bridge connections found in the persistent database. Unfortunately, there is no way to recognize bridge connections at that point. It would require an additional flag in the database.
- When a bridge connection is being configured and a context for the client ID already existed, remove the corresponding session expiration timer.
- Don't write bridge connections to the persistent database. I'm not completely sure but I think those records are useless for bridge connections anyway (although apparently their existence in the persistent database was a conscious decision).
- Ignore bridge connections when checking session expiration timers.

I have decided to go with option number 2.

### DIFF
--- a/src/bridge.c
+++ b/src/bridge.c
@@ -84,6 +84,7 @@ int bridge__new(struct mosquitto__bridge *bridge)
 	if(new_context){
 		/* (possible from persistent db) */
 		mosquitto__free(local_id);
+		session_expiry__remove(new_context);
 	}else{
 		/* id wasn't found, so generate a new context */
 		new_context = context__init(INVALID_SOCKET);


swiergot:
Nevermind, apparently the fix (https://github.com/eclipse/mosquitto/commit/6d240a9d183413fca7d2a4f42eae803ba84af97c) has been on the fixes branch for over 3 months. There just hasn't been any new release since.

PR #2906, opened 2023-09-25
Perform ACK completion after the callback to avoid message loss

### LINKED ISSUE #2884: Client PUB completion (PUBACK, PUBREC/PUBCOMP) should be performed after the application callback.
The current client implementation will send the PUBACK (likely the same for PUBREC/COMP) _before_ calling the application callback. This can result in messages being lost if the application crashes during the callback.

According to the V3/V5 specs Section 4.3, I believe the intent of the protocol is for the ACK/COMP messages to indicate that the _application_ takes ownership of the message, and not the protocol library.

I propose moving the message acknowledgment _after_ the callback for QoS 1 and 2.

------

### Details

Currently, `send__puback` is called before the `on_message` callback: https://github.com/eclipse/mosquitto/blob/1ed275b1d4f49cdeca45bbcaa605e75fafa1b140/lib/handle_publish.c#L142

https://github.com/eclipse/mosquitto/blob/1ed275b1d4f49cdeca45bbcaa605e75fafa1b140/lib/handle_publish.c#L146

If the ordering is reversed (`on_message` before `send__puback`), even if the application crashes in the callback the message isn't lost and will be sent again by the broker after the application recovers (reconnects).

_Version_: Mosquitto Client (master branch pulled around tag 2.0.16, built from sources).

Related: #118

CIPop:

- [X] Have you successfully run `make test` with your changes locally? (Ran `ctest -E broker` as broker baseline testing for `develop` hangs.)

-----

### Abstract

The current Mosquitto client implementation might lose QoS1 and QoS2 messages if the application crashes during the hand-off callback.

Based on [MQTT section 4.4 Message retry delivery](https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901238), the application (not the MQTT library) takes ownership of the message:
- For QoS 1: _before_ sending the PUBACK.
- For QoS 2: _before_ sending the PUBREC

### Behavior changes

The current implementation, on receiving a PUB:
- QoS 1: sends the PUBACK, __calls the callback__
- QoS 2: sends PUBREC, receives PUBREL, sends PUBCOMP, __calls the callback__

With this change, on receiving a PUB:
- QoS 1: __calls the callback__, sends the PUBACK
- QoS 2: __calls the callback__, sends PUBREC, receives PUBREL, sends PUBCOMP

### Notes
- The reordering of ACKs _might_ be a breaking change for certain applications. If the callback crashes the client, the callback will be called more than once, with the same message (even for QoS2), until both callback execution and ACK sending is successful.
- The reordering allows applications to crash while processing the callback without losing the message: the broker is supposed to re-send a QoS1 message after the subscriber reconnects, if a PUBACK was not received. Similarly, the broker is supposed to re-send a QoS2 message after the subscriber reconnects if a PUBREC was not received.
- Because the library does not expose a receive ACK callback (PUBACK, PUBREC, etc), the application would not know when these acknowledgement messages are sent. The message loop must run for a while to ensure all ACKs have been drained from the queue. (See test changes for details.)
  
Fixes #2884

### DIFF
--- a/lib/handle_publish.c
+++ b/lib/handle_publish.c
@@ -156,14 +156,19 @@ int handle__publish(struct mosquitto *mosq)
 			return MOSQ_ERR_SUCCESS;
 		case 1:
 			util__decrement_receive_quota(mosq);
-			rc = send__puback(mosq, mid, 0, NULL);
+			// Application callbacks are called before PUBACK is sent (and ownership given to the 
+			// receiver). An application crash within the callback will result in retransmission.
 			callback__on_message(mosq, &message->msg, properties);
+			rc = send__puback(mosq, mid, 0, NULL);
 			message__cleanup(&message);
 			mosquitto_property_free_all(&properties);
 			return rc;
 		case 2:
 			message->properties = properties;
 			util__decrement_receive_quota(mosq);
+			// Application callbacks are called before PUBREC is sent (and ownership given to the 
+			// receiver). An application crash within the callback will result in retransmission.
+			callback__on_message(mosq, &message->msg, message->properties);
 			rc = send__pubrec(mosq, mid, 0, NULL);
 			pthread_mutex_lock(&mosq->msgs_in.mutex);
 			message->state = mosq_ms_wait_for_pubrel;
--- a/lib/handle_pubrel.c
+++ b/lib/handle_pubrel.c
@@ -112,9 +112,7 @@ int handle__pubrel(struct mosquitto *mosq)
 
 	rc = message__remove(mosq, mid, mosq_md_in, &message, 2);
 	if(rc == MOSQ_ERR_SUCCESS){
-		/* Only pass the message on if we have removed it from the queue - this
-		 * prevents multiple callbacks for the same message. */
-		callback__on_message(mosq, &message->msg, message->properties);
+		// Message was already passed to the application prior to PUBREC.
 		message__cleanup(&message);
 	}else if(rc == MOSQ_ERR_NOT_FOUND){
 		return MOSQ_ERR_SUCCESS;
--- a/test/lib/11-prop-recv-qos1.py
+++ b/test/lib/11-prop-recv-qos1.py
@@ -21,8 +21,10 @@ def do_test(conn, data):
     mosq_test.do_receive_send(conn, connect_packet, connack_packet, "connect")
 
     conn.send(publish_packet)
-    mosq_test.expect_packet(conn, "puback", puback_packet)
+    # Application sends the "ok" publish within the callback.
     mosq_test.expect_packet(conn, "ok", ok_packet)
+    # After the callback exits, the puback is sent.
+    mosq_test.expect_packet(conn, "puback", puback_packet)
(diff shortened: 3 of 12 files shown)


lib/handle_publish.c
```diff
@@ -156,14 +156,19 @@ int handle__publish(struct mosquitto *mosq)
 			return MOSQ_ERR_SUCCESS;
 		case 1:
 			util__decrement_receive_quota(mosq);
-			rc = send__puback(mosq, mid, 0, NULL);
+			// Application callbacks are called before PUBACK is sent (and ownership given to the
```
ericwol-msft, line 159: should be using /* */ for comments

ericwol-msft, line 159:
```suggestion
			// Application callbacks should be called before PUBACK is sent (and ownership given to the 
```

lib/handle_publish.c
```diff
@@ -156,14 +156,19 @@ int handle__publish(struct mosquitto *mosq)
 			return MOSQ_ERR_SUCCESS;
 		case 1:
 			util__decrement_receive_quota(mosq);
-			rc = send__puback(mosq, mid, 0, NULL);
+			// Application callbacks are called before PUBACK is sent (and ownership given to the 
+			// receiver). An application crash within the callback will result in retransmission.
```
ericwol-msft, line 160:
```suggestion
			// receiver). An application crash within the callback will result in message retransmission.
```

lib/handle_publish.c
```diff
@@ -156,14 +156,19 @@ int handle__publish(struct mosquitto *mosq)
...
 			callback__on_message(mosq, &message->msg, properties);
+			rc = send__puback(mosq, mid, 0, NULL);
 			message__cleanup(&message);
 			mosquitto_property_free_all(&properties);
 			return rc;
 		case 2:
 			message->properties = properties;
 			util__decrement_receive_quota(mosq);
+			// Application callbacks are called before PUBREC is sent (and ownership given to the
```
ericwol-msft, line 169:
```suggestion
			// Application callbacks should be called before PUBREC is sent (and ownership given to the 
```

lib/handle_publish.c
```diff
@@ -156,14 +156,19 @@ int handle__publish(struct mosquitto *mosq)
...
 			callback__on_message(mosq, &message->msg, properties);
+			rc = send__puback(mosq, mid, 0, NULL);
 			message__cleanup(&message);
 			mosquitto_property_free_all(&properties);
 			return rc;
 		case 2:
 			message->properties = properties;
 			util__decrement_receive_quota(mosq);
+			// Application callbacks are called before PUBREC is sent (and ownership given to the 
+			// receiver). An application crash within the callback will result in retransmission.
```
ericwol-msft, line 170:
```suggestion
			// receiver). An application crash within the callback will result in message retransmission.
```

ericwol-msft (approved):

RLeclair (approved):

PR #2941, opened 2023-10-30
Unix Domain Sockets support for websockets listener

okisseloff:
### Description

It looks like mosquitto listeners support Unix Domain Sockets to serve mqtt over TCP, while using them with websockets fails.  
`libwebsockets` has a compile option named `LWS_UNIX_SOCK` that enables `UDS` support.
Suggested change uses the same configuration syntax for websockets listener:
```
listener 0 /path/to/unix.sock
protocol websockets
```
but obtains the file name from another `unix_socket_path` listener property.
It also enables UDS support for `lws_context` instance by appending `LWS_SERVER_OPTION_UNIX_SOCK` options flag.

### Note

I had tried the version of the patch that used `LWS_USE_UNIX_SOCK` definition to test if `libwebsockets` support UDS, but it should work only when compiling, while at runtime the dynamic library could be replaced with another instance, so I thought it doesn't make any sense. Haven't found any runtime test for UDS support in `libwebsockets`.

-----

### DIFF
--- a/src/websockets.c
+++ b/src/websockets.c
@@ -688,8 +688,13 @@ void mosq_websockets_init(struct mosquitto__listener *listener, const struct mos
 
 	memset(&info, 0, sizeof(info));
 	if(listener->port == 0){
+#ifdef WITH_UNIX_SOCKETS
 		info.iface = listener->unix_socket_path;
 		info.options |= LWS_SERVER_OPTION_UNIX_SOCK;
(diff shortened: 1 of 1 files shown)


ralight:
Thank you for this contribution, however I don't wish to add further features to the code making use of libwebsockets as this is replaced in the develop branch.

PR #2969, opened 2023-12-18
Copy scope id when binding to IPv6 Link-Local address

### LINKED ISSUE #2696: bind_interface doesn't work, mosquitto only binds to localhost
mosquitto-2.0.15 on Gentoo Lignux here. `strace` doesn't tell any hints on why doesn't it work. `listener` works but is unsuitable for dynamic IP addresses.
```
∞ grep -Ev '^#|^$' /etc/mosquitto/mosquitto.conf 
persistent_client_expiration 1y
bind_interface eth2
autosave_interval 1800
persistence true
persistence_file mosquitto.db
persistence_location /var/lib/mosquitto/
log_dest syslog
log_type information
allow_anonymous true
```

ThymoNL:

-----

Configuring a listener with bind_interface on an interface with an ipv6 link-local address causes Mosquitto to exit with "Error: invalid argument". This happens due to the the IPv6 scope ID not being copied to the returned addrinfo struct when resolving interfaces matching `bind_interface`.

This PR adds a commit that also copies the sin6_scope_id when a matching interface is found.

Fixes #2696

### DIFF
--- a/src/net.c
+++ b/src/net.c
@@ -673,6 +673,9 @@ static int net__bind_interface(struct mosquitto__listener *listener, struct addr
 						memcpy(&((struct sockaddr_in6 *)rp->ai_addr)->sin6_addr,
 								&((struct sockaddr_in6 *)ifa->ifa_addr)->sin6_addr,
 								sizeof(struct in6_addr));
+
+						((struct sockaddr_in6 *)rp->ai_addr)->sin6_scope_id = ((struct sockaddr_in6 *)ifa->ifa_addr)->sin6_scope_id;
+
 						freeifaddrs(ifaddr);
 						return MOSQ_ERR_SUCCESS;
 					}
--- a/test/broker/readme.txt
+++ b/test/broker/readme.txt
@@ -17,3 +17,5 @@ Numbering is as follows:
 10: Listener tests
 11: Persistence tests
 12: Property tests
+13: Malformed tests
+14: Dynamic security tests


ThymoNL:
@ralight May I gently bump this your way?  Hoping you can take a look at this bugfix. 🙂

src/net.c
```diff
@@ -673,6 +673,9 @@ static int net__bind_interface(struct mosquitto__listener *listener, struct addr
 						memcpy(&((struct sockaddr_in6 *)rp->ai_addr)->sin6_addr,
 								&((struct sockaddr_in6 *)ifa->ifa_addr)->sin6_addr,
 								sizeof(struct in6_addr));
+
+						((struct sockaddr_in6 *)rp->ai_addr)->sin6_scope_id = ((struct sockaddr_in6 *)ifa->ifa_addr)->sin6_scope_id;
```
ralight, line 677: Can you explain why the memcpy call in line 673 isn't sufficient to copy this value already?

ThymoNL: Is your question why only copying the in6_addr is not enough? That's because for link-local addresses they are only valid for a specific interface, you can specify which interface the address belongs to with the scope id.

If your question is why I didn't group everything into a single memcpy call. That was for readability.

ThymoNL:
@ralight Can I do anything else to answer your questions in order to get this merged? :)

ThymoNL:
@ralight friendly reminder 🙂

ralight:
@ThymoNL Thanks for the reminder. I would be happy to merge this, however I'd be much happier if there was a test to check the link local case. Any ideas on adding one?

ralight:
Having looked into this a bit more, I think it's going to be impractical to write a general test for it.

The github runners do get a link local address:

```
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 60:45:bd:4a:dd:40 brd ff:ff:ff:ff:ff:ff
    inet 10.1.0.23/16 metric 100 brd 10.1.255.255 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::6245:bdff:fe4a:dd40/64 scope link 
       valid_lft forever preferred_lft forever
```

However in other places, like my home, I get other non-link local IPv6 addresses and the bug does not manifest  there in the same way.

ThymoNL:
Yeah, I remember looking at this back then and not finding a trivial way to write a test either. Thanks for merging!

PR #3018, opened 2024-03-12
Add Timezone data to dockerfile

ajaypala:
Adding timezone data (tzdata) package to dockerfile to allow Mosquitto to use local time in logs instead of UTC time.

Very simple change. Another developer can copy the changes and pull/merge it themselves.


ralight:
Thank you for the PR. It isn't quite right because you're adding to the build-deps which later get removed. I've now committed a change for all Dockerfiles that includes tzdata in the right place.

Issue #2990, opened 2024-02-01
pkg-config issue on Raspberry Pi bullseye


maxim-sloyko-yohana:
I'm not sure this is the right place to report the packaging problem, so my apologies if it's not.
The `libmosquitto.pc` file in Raspberry Pi bullseye distribution currently looks like this:

```ini
prefix=/usr
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/lib

Name: mosquitto
Description: mosquitto MQTT library (C bindings)
Version: 2.0.11
Cflags: -I${includedir}
Libs: -L${libdir} -lmosquitto
```

However, `libmosquitto.so` is actually installed at `/usr/lib/aarch64-linux-gnu/libmosquitto.so`.
This can lead to `-lmosquitto not found` linking error. On my system I had to change the `libdir=` line above to
```ini
libdir=${exec_prefix}/lib/aarch64-linux-gnu
```
for my `pkg-config` based build to work.

ralight:
This is fixed in https://github.com/eclipse-mosquitto/mosquitto/commit/c34be255d30800459a285b0a7e31547683064cad and https://github.com/eclipse-mosquitto/mosquitto/commit/ab4ac42b9043eeab0fa1e3fe2f4a3cf8ac402ab3

PR #3027, opened 2024-04-01
Fix libdir location in pkgconfig files.

richmattes:
The pkgconfig files currently hardcode the library directory as` ${prefix}/lib`, but the library installation directory in `config.mk` is set as `${prefix}/lib${LIB_SUFFIX}`.  This causes a mismatch between the library installation directory and the pkgconfig file on redhat-based 64-bit systems.

This commit remedies the issue for both the Makefile and CMake build by modifying the pkgconfig files to use `CMAKE_INSTALL_LIBDIR` instead of `lib`.  In the CMake system, this directly reflects the installation path.  The Makefiles were modified to replace `CMAKE_INSTALL_LIBDIR` with `lib${LIB_SUFFIX}`, matching the `libdir` definition in `config.mk` and following the pattern used in defining the `prefix`.

-----

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -75,7 +75,7 @@ endif
 	$(INSTALL) ../include/mqtt_protocol.h "${DESTDIR}${incdir}/mqtt_protocol.h"
 	$(INSTALL) -d "${DESTDIR}${libdir}/pkgconfig"
 	$(INSTALL) -m644 ../libmosquitto.pc.in "${DESTDIR}${libdir}/pkgconfig/libmosquitto.pc"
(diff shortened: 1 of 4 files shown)


richmattes:
This should also address #2990.

ralight:
Thank you, this is merged in https://github.com/eclipse-mosquitto/mosquitto/commit/c34be255d30800459a285b0a7e31547683064cad and https://github.com/eclipse-mosquitto/mosquitto/commit/ab4ac42b9043eeab0fa1e3fe2f4a3cf8ac402ab3

Issue #2944, opened 2023-11-07
mosquitto_passwd.c: problematic use of OPENSSL_API_COMPAT


dirkfeytons:
According to https://www.openssl.org/docs/man3.1/man7/OPENSSL_API_COMPAT.html the OPENSSL_API_COMPAT macro is expected to be set by the user on the compiler command line or at the top of a source file before any OpenSSL header is included.
The way it's used at https://github.com/eclipse/mosquitto/blob/3923526c6b4c048bbecad2506c4c9963bc46cd36/apps/mosquitto_passwd/mosquitto_passwd.c#L459 can lead to an undesired build error: since OPENSSL_API_COMPAT is not defined in a standard Mosquitto build, the condition `OPENSSL_API_COMPAT < 0x10100000L` in the code will evaluate to true and the call to `	OpenSSL_add_all_digests()` is included. However, when OpenSSL is built with `no-deprecated` then this symbol does not exist and the build fails.

The check should simply be removed from the code; only checking for `OPENSSL_VERSION_NUMBER < 0x10100000L` is sufficient.

Issue #2907, opened 2023-09-28
Can't establish secure SSL TLS v1.2 connection among mosquitto mqtt broker and the client


SamShekhovtsov:
The mosquitto mqtt broker (latest verison available) is configured to use TLS v1.2, it is verified that the server works fine using the TLS encryption with live clients like the following: 
https://testclient-cloud.mqtt.cool/
MQTT-Explorer           http://mqtt-explorer.com/
The client application is setup to support the TLS v1.2 encryption. The client application can connect to the live MQTT brokers like AWS Amazon MQTT broker and the https://www.hivemq.com/ broker. Everything works fine with the TLS v1.2 encryption, that data exchange is protected. 
When trying to pair the mosquitto mqtt broker and the client application - the TLS handshake failure is experienced:

```
OpenSSL Error[0]: error:140370E5:SSL routines:ACCEPT_SR_KEY_EXCH:ssl handshake failure
```
when I try to debug the SSL/TLS using the Wireshark, I see that there is a Client Hello, after that the Server Hello, then Certificate, Server Key Exchange, Server Hello Done and it stops at that point.

After the deep investigation, it looks like the mosquitto MQTT broker is using the LibreSSL (libssl) library and not the OpenSSL (one of the forks) as it stated. 
There are two possibilties: (1) server is indeed sending something bad in the ServerKeyExchange to this client, but other clients don't see the problem because they don't use the DHE (at all), don't use the same type of DHE (FF vs EC), or don't use the same curve/group for DHE, or get a different signature type. (2) server is sending valid ServerKeyExchange that triggers a bug in client, but other servers don't because they don't use DHE, don't use the same DHE, or the same curve/group, or the same signature.

Issue #2779, opened 2023-04-03
Update openSSL to v3.x


olgapuig:
Eclipse Mosquittov2.0.15 uses openSSLv1.1.1q and it is close to the [End Of Life ](https://www.openssl.org/blog/blog/2023/03/28/1.1.1-EOL/#:~:text=OpenSSL%201.1.1%20was%20released%20on%2011th%20September%202018%2C,receiving%20publicly%20available%20security%20fixes%20after%20that%20date).

openSSL1.1.1q is not the latest version for the 1.1.1 branch, it is 1.1.1u.  There have been multiple security vulnerabilities reported and fixed in the versions between 1.1.1q and 1.1.1u.  The [change log from OpenSSL](https://www.openssl.org/news/cl111.txt)

This is a request to update to openSSLv3.x due to Cybersecurity. 

Note: The latest stable version is the openSSLv3.1 series supported until 14th March 2025. Also available is the 3.0 series which is a Long Term Support (LTS) version and is supported until 7th September 2026. The previous LTS version (the 1.1.1 series) is also available and is supported until 11th September 2023.

ralight:
Good points.

gonzabrusco:
Hi @ralight if I install Mosquitto 2.0.15 on Ubuntu 22.04 that comes with OpenSSL 3.0.2 ... this means it's incompatible and I should install an older version of Ubuntu? 

I'm having a problem with some devices trying to connect Mosquitto 2.0.15. The problem is that they can't achieve the connection using TLS 1.1 (the devices are deployed and impossible to update). Even setting tls_version tlsv1.1 on the config file I get the following:

```
2023-04-27T_14:40:05: New connection from 181.47.142.45:6724 on port 8883.
2023-04-27T_14:40:05: OpenSSL Error[0]: error:0A000102:SSL routines::unsupported protocol
2023-04-27T_14:40:05: Client <unknown> disconnected: Protocol error.
2023-04-27T_14:40:07: Client connection from 181.47.142.45 failed: error:0A000102:SSL routines::unsupported protocol.
2023-04-27T_14:40:08: Client connection from 181.47.142.45 failed: error:0A000102:SSL routines::unsupported protocol.
```

Sorry for piggybacking this issue report but I think this is related to the problem I'm having.

gonzabrusco:
[...]

PR #3050, opened 2024-05-10
Fix multiple build warnings (raises cmake_minimum_required and sets OPENSSL_API_COMPAT)

carstenandrich:

-----

This PR fixes multiple build warnings I stumbled upon integrating libmosquitto into a project of mine.
Changes are split into 4 separate commits that address individual issues, to enable cherry picking:
1. raise cmake_minimum_required(VERSION 3.1) to 3.5 to resolve a [deprecation warning with CMake 3.27 and above](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings). CMake 3.5 is from April 2016.
2. fix unused parameter warning in mosquitto_void_option() that only occurs when building without TLS support
3. fix unused variable warning in net__socket_connect_tls() by removing 2 entirely unused variables
4. set OpenSSL API version compatibilty macro [OPENSSL_API_COMPAT](https://www.openssl.org/docs/manmaster/man7/OPENSSL_API_COMPAT.html) to OpenSSL 1.1. currently mosquitto uses the ENGINE API, which was deprecated with OpenSSL 3.0, causing deprecation warnings when compiling with GCC: `src/net.c:566:9: warning: ‘ENGINE_free’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]`

Nr. 4 is related to #2944 and #2779. In the latter it was [suggested to define `OPENSSL_NO_ENGINE` to silence the warning](https://github.com/eclipse/mosquitto/issues/2779#issuecomment-1990976800), but for lack of documentation on `OPENSSL_NO_ENGINE` and its possible side effects, setting `OPENSSL_API_COMPAT` appears to be the cleaner approach. However, [migrating from OpenSSL ENGINE to Providers API](https://www.openssl.org/docs/manmaster/man7/migration_guide.html#Engines-and-METHOD-APIs) is probably the only long-term solution to address the TLS handshake issues in #2779 and possibly #2907.


ralight:
This all looks good, thank you.

PR #3054, opened 2024-05-20
OpenSSL: fix spurious SSL connection aborts

koranyellow:
### Description

Was seeing spurious SSL connection aborts using libmosquitto and OpenSSL. I tracked it down to uncleared error state on the OpenSSL error stack - patch attached deals with that.

### Rough idea of problem:

Code that uses libmosquitto calls some library that uses OpenSSL but don't clear the OpenSSL error stack after an error. lib/net_mosq.c calls SSL_read which eventually gets an EWOULDBLOCK from the OS. Returns -1 to indicate an error lib/net_mosq.c calls SSL_get_error. First thing, SSL_get_error calls ERR_get_error to check the OpenSSL error stack, finds an old error and returns SSL_ERROR_SSL instead of SSL_ERROR_WANT_READ or SSL_ERROR_WANT_WRITE.

lib/net_mosq.c returns an error and aborts the connection

### Solution:
Clear the openssl error stack before calling SSL_* operation if we're going to call SSL_get_error afterwards.

### Notes:
This is much more likely to happen with multi because it's easier to intersperse other calls to the OpenSSL library in the same thread.

-----

- [-] If you are contributing a new feature, is your work based off the develop branch?

-----

The problem appears after [#commit.](https://github.com/eclipse/mosquitto/commit/47dadb902dacb4d963ba9ec34803f62ea418e964)

### DIFF
--- a/lib/net_mosq.c
+++ b/lib/net_mosq.c
@@ -976,6 +976,7 @@ ssize_t net__read(struct mosquitto *mosq, void *buf, size_t count)
 	errno = 0;
 #ifdef WITH_TLS
 	if(mosq->ssl){
+		ERR_clear_error();
 		ret = SSL_read(mosq->ssl, buf, (int)count);
 		if(ret <= 0){
 			ret = net__handle_ssl(mosq, ret);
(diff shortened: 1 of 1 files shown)


ralight:
Thank you, I've added this change to a separate commit because you haven't signed the ECA - but there is no other way to implement this change. It will be in 2.0.19.

PR #3090, opened 2024-07-31
fix CLOCK_BOOTTIME use on kernels not having it

### LINKED ISSUE #3089: Failure with kernels not supporting CLOCK_BOOTTIME
I need to run mosquitto (client using libmosquitto) on a platform which has a Linux kernel without `CLOCK_BOOTTIME` but where the libc headers declare `CLOCK_BOOTTIME`, since they are decoupled from the kernel itself.

Under these circumstances libmosquitto goes crazy and fails with `Keepalive exceeded` errors. I traced the issue to `mosquitto_time()` that will do `clock_gettime(CLOCK_BOOTTIME, &tp)` if `CLOCK_BOOTTIME` is defined by the libc headers, without checking the return value. Since on this platform there is no `CLOCK_BOOTTIME` the call returns `-EINVAL` and the value returned by `mosquitto_time()` is basically garbage, so any timestamp comparisons go haywire.

The fix is relatively simple, and basically boils down to probing, at runtime, for `CLOCK_BOOTTIME` and using `CLOCK_MONOTONIC` as fallback.

Mosquitto version: 2.0.18

diego-santacruz:
When the libc headers used to build mosquitto define CLOCK_BOOTTIME but the kernel where mosquitto runs does not implement CLOCK_BOOTTIME, all timestamps are wrong (uninitialized data), because there is no check for the success of the clock_gettime() calls.

This adds probing for the availability of CLOCK_BOOTTIME from mosquitto_lib_init(), falling back to CLOCK_MONOTONIC, and then modifies mosquitto_time() to use the selected clock. Probing at init time avoids having to do two clock_gettime() calls for every timestamp if CLOCK_BOOTTIME is not available.

It also fixes a similar problem in mosquitto_lib_init().

This fixes #3089 

-----

### DIFF
--- a/lib/mosquitto.c
+++ b/lib/mosquitto.c
@@ -38,6 +38,7 @@ Contributors:
 #include "mqtt_protocol.h"
 #include "net_mosq.h"
 #include "packet_mosq.h"
+#include "time_mosq.h"
 #include "will_mosq.h"
 
 static unsigned int init_refcount = 0;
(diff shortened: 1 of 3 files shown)


diego-santacruz:
I have amended the PR as I had mistakenly used the `clock_t` type instead of `clockid_t` (the compiler had not given any warning about this mismatch).

ralight:
Looks good to me, thank you.

PR #3122, opened 2024-10-04
Let systemd create directories

gchamp20:
### Context

Manually creating directories conflicts with systemd service hardening features. Particularly, `ProtectSystem=strict` mounts the filesystem has read-only for the processes started by the unit which leads to `mkdir` failing.

By setting `User=mosquitto` and adding `RuntimeDirectory` and `LogsDirectory`, systemd creates `/run/mosquitto` and `/var/log/mosquitto` with the right permissions even `ProtectSystem=strict` is used.

Adding `User=mosquitto` also has the side effect of running the daemon as the user `mosquitto`. I 

### Checklist

https://accounts.eclipse.org/users/gchamp20, submitted, appears to be pending? I can still re-submit but I now get an error.

### DIFF
--- a/service/systemd/mosquitto.service.notify
+++ b/service/systemd/mosquitto.service.notify
@@ -5,15 +5,14 @@ After=network.target
 Wants=network.target
 
 [Service]
+User=mosquitto
 Type=notify
 NotifyAccess=main
 ExecStart=/usr/sbin/mosquitto -c /etc/mosquitto/mosquitto.conf
 ExecReload=/bin/kill -HUP $MAINPID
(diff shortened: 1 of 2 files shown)


jspngh:
This solves an issue that I'm having with the `/run/mosquitto/` directory being inaccessible to members of the mosquitto group.
The `740` permissions should be `750`: without execute permission, you can't enter the directory.

Systemd uses `755`, but I guess that should be okay?
@ralight, do you agree with this change?

ralight:
Thanks for the explanation, I think this sounds like a sensible idea.

PR #3150, opened 2024-10-31
Fix pkgconfig libdir

nullr0ute:

-----


ralight:
I understand the rationale, however at the moment the pure Makefile approach also installs the .pc files. This means there needs to be a string replacement when installing from Makefile otherwise it's completely broken there. Is there a better way of detecting the lib64 component?

nullr0ute:
> I understand the rationale, however at the moment the pure Makefile approach also installs the .pc files. This means there needs to be a string replacement when installing from Makefile otherwise it's completely broken there. Is there a better way of detecting the lib64 component?

I missed that the Makefile approach stole CMake vars in that approach, will update the patch so both ways are fixed.

ralight:
This is fixed in https://github.com/eclipse-mosquitto/mosquitto/commit/c34be255d30800459a285b0a7e31547683064cad and https://github.com/eclipse-mosquitto/mosquitto/commit/ab4ac42b9043eeab0fa1e3fe2f4a3cf8ac402ab3

nullr0ute:
Awesome, thanks for following this up, it slipped off my radar.

PR #3158, opened 2024-11-05
Add systemd watchdog support

gportay:
Add possibility to send keep-alive ping to systemd watchdog. Also update notify service file to ping service manager every 3 minutes; mosquitto pings actually every 1 minute and 30 seconds (half period).

### DIFF
--- a/service/systemd/mosquitto.service.notify
+++ b/service/systemd/mosquitto.service.notify
@@ -6,6 +6,7 @@ Wants=network.target
 
 [Service]
 Type=notify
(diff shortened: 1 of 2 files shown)


gportay:
Dear maintainers, is there any chance to get some review? Thanks in advance.

ralight:
Thanks very much for this, I've reworked the patch for the `develop` branch and pushed it. Github doesn't recognise it as this PR of course, so I have to close instead.

PR #3200, opened 2025-01-14
[JET-13007] Fix bridge

ricklvt:
## Purpose
Mosquitto appears to be using the `session_expiry` setting for two separate purposes.  The first one, which matches the spec is to keep messages not yet delivered to a client "session" for the requested amount of time.  The second one, which I think is probably a bug in mosquitto is to close any connections that have been up for `session_expiry` seconds.  I plan to followup with the mosquitto project to inform them that I think this is a misinterpretation of how `session_expiry` should be used.  But in the meantime, this change in our (LVT) fork will prevent it from closing bridge connections.  I've also added some additional logging to help debug if we still see problems (it's possible there is more than one bug we need to fix).

## Review instructions
Select the option to "Hide whitespace" changes.

## Testing
I was able to reproduce the session stalling problem on my test unit (mobile 6), and then confirm that this change fixes it.   In order to repro the problem, I had to change the `session_expiry` setting in `/etc/mosquitto/conf.d/02_bridge.conf`, but that wasn't enough.  I also had to remove `/var/lvt/mosquittodb/mosquitto.db` before restarting, because otherwise mosquitto reloads the `session_expiry` value from there instead of using the new value.

-----

### DIFF
--- a/src/bridge.c
+++ b/src/bridge.c
@@ -245,17 +245,18 @@ int bridge__connect_step1(struct mosquitto *context)
 		}
 	}
 
-	log__printf(NULL, MOSQ_LOG_NOTICE, "Connecting bridge (step 1) %s (%s:%d)", context->bridge->name, context->bridge->addresses[context->bridge->cur_address].address, context->bridge->addresses[context->bridge->cur_address].port);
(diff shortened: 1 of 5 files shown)


ricklvt:
this was meant to be a PR in our fork.  my bad.

Issue #3201, opened 2025-01-14
Python 3.13 - Test fails with "Basic Constraints of CA cert not marked critical"


narc-Ontakac2:
See the [Debian bug 1092750](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1092750) for details.

PR #3203, opened 2025-01-15
Fixed tests failing with python 3.13 (Closes: #3201)

narc-Ontakac2:
Python 3.13 is more strict concerning what it accepts as root certificates. This PR adapts the openssl.conf to fix this and regenerates the certificates using gen.sh.


ralight:
Thanks very much, I've not got any systems with Python 3.13 yet so hadn't seen this.

PR #3227, opened 2025-03-07
fix: compile failed when WITH_TLS is off

worry80:
when WITH_TLS is off, errors such as

mosquitto/src/conf.c:2318:54: error: ‘struct mosquitto__listener’ has no member named ‘cafile’
 2318 |                  bool cafile = !!config->listeners[i].cafile;

may occur during the compilation process


pfichtner:
Had this compile error, fix works for me

Issue #3239, opened 2025-03-16
man mosquitto_ctrl needs a linebreak


narc-Ontakac2:
```
W: mosquitto: groff-message troff:<standard input>:36: warning [p 1, 2.3i]: cannot break line [usr/share/man/man1/mosquitto_ctrl.1.gz:2]
```
A patch will follow.

rzr:
LGTM

Relate-to: https://salsa.debian.org/debian-iot-team/mosquitto/-/merge_requests/29

narc-Ontakac2:
Should be done, see PR #3240 .

PR #3240, opened 2025-03-17
Added line breaks to man mosquitto_ctrl.1 (Closes: #3239)

narc-Ontakac2:


ralight:
Thank you for this, I like the idea of making the man pages easier to read. I've taken this idea and grouped the options into different sections which I think really helps. As such I'm going to close this PR, but thank you for the inspiration.

PR #3283, opened 2025-06-07
log: show process user in console

atifali:

-----

This is a simple addition to the mosquitto binary to be able to show the user that the process is being run under since there are places where the user privileges are being changed depending on what the mosquitto process is ran under. This makes debugging easier by showing the current process user name in the log. If its an unknown user the program aborts.


Daedaluz:
why would having an unknown uid -> username be a fatal error?
can't it just print the id instead?

atifali:
@Daedaluz doesn't a username of `unknown` mean something is not quite right with the configuration of the user? In which case we would want that fixed before that config is used to run mosquitto. I could be wrong here though! @ralight thoughts on this?

Daedaluz:
perhaps the usecase is narrow, but I think there are some cases where the user / group / privileges are set before the software starts in some container environment configs.

to clarify what i meant; it doesn't have to only print the id, it could fall back to just print the id if the user isn't found.

ralight:
@atifall I think this is a good idea to include, thank you. I do agree with @Daedaluz though, there isn't a hard requirement for there to be a username mapped to a uid. In that case it would be fine to print the uid instead.

atifali:
@ralight I believe this should be good to merge now. Let me know if there are any other changes needed here.

ralight:
Thank you. It still required a fix for Windows, but I've done that. I've merged it in https://github.com/eclipse-mosquitto/mosquitto/commit/9f1e876d0eaa2d209ae7092d7822c3f36c1dc528, but github hasn't picked it up here for some reason. I'll have to close it manually. Thanks again!

PR #3307, opened 2025-07-10
Add migration of mosquitto.conf to DynSec migration script

SteveHamann:

-----

### DIFF
--- a/plugins/dynamic-security/migrate_to_dynsec.py
+++ b/plugins/dynamic-security/migrate_to_dynsec.py
@@ -418,16 +418,71 @@ def migrate_to_dynsec(
     return dynsec_config
 
 
+def migrate_mosquitto_conf(
+    mosquitto_conf: str, dynsec_lib_path: Path, dynsec_config_path: Path
+) -> str:
+    migrated_mosquitto_conf: list[str] = []
+    for line in mosquitto_conf.splitlines():
+        if line.startswith("acl_file"):
+            continue
+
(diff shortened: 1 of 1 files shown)


ralight:
This is good - but could we also load the acl and password files for migration when parsing the config file? So a migration would be `migrate_to_dynsec.py --conf mosquitto.conf --dynsec-lib /path` and everything would be done?

plugins/dynamic-security/migrate_to_dynsec.py
```diff
@@ -418,16 +418,50 @@ def migrate_to_dynsec(
...
         "--pw-file", required=True, type=str, help="Path to password file"
     )
+    parser.add_argument(
+        "--conf", required=True, type=str, help="Path to mosquitto.conf file"
+    )
+    parser.add_argument(
+        "--dynsec-lib",
+        required=True,
+        type=str,
+        help="Path to mosquitto_dynamic_security.so",
```
ralight, line 457: Could this please be "Path to mosquitto_dynamic_security.so/.dll"

ralight (commented):
You've done exactly what I suggested so this is entirely my fault - sorry. I think that having the old behaviour and new behaviour at the same time is useful. In other words having --conf autodetect the acl/password files, but still having the --pw-file and --acl-file options present to just migrate those files if needed. Would that be ok?

ralight:
Thanks Steve

PR #3321, opened 2025-07-23
Fix CMake builds with WITH_TLS=OFF

### LINKED ISSUE #3318: Build without SSL issues: Target "libmosquitto" links to OpenSSL::SSL but the target was not found
Hello,

I am trying to build without any support of TLS and OpenSSL and I have errors on the cmake command.

It seems to failed to found OpenSSL::SSL as libraries are not installed on my machine.

```
CMake Error at external/mosquitto/lib/CMakeLists.txt:103 (target_link_libraries):
  Target "libmosquitto" links to:

    OpenSSL::SSL

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at external/mosquitto/lib/CMakeLists.txt:131 (target_link_libraries):
  Target "libmosquitto_static" links to:

    OpenSSL::SSL

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



CMake Error at external/mosquitto/src/CMakeLists.txt:214 (target_link_libraries):
  Target "mosquitto" links to:

    OpenSSL::SSL

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

However I set the WITH_TLS=OFF and WITH_TLS_PSK=OFF so I would expect not to build with OpenSSL support.

Thus I checked inside the lib/CMakeLists.txt and src/CMakeLists.txt and I saw that OpenSSL library is declared all the time as dependency of mosquittto.

So I commented line 166 of src/CMakeLists.txt saying: `set (MOSQ_LIBS ${MOSQ_LIBS} OpenSSL::SSL)`

And I replaced line 63 of lib/CMakeLists.txt from : `set (LIBRARIES OpenSSL::SSL)` to `set (LIBRARIES "")`

I managed to successfully rebuild the following targets without any error on Rocky 9 64 bits:
- libmosquitto
- libmosquittopp
- mosquitto
- mosquitto_pub
- mosquitto_sub

Can you fix this in your cmake ?

I would suggest modifiying src/CMakeLists.txt this way

```cmake
if (WITH_TLS)
    set (MOSQ_LIBS ${MOSQ_LIBS} OpenSSL::SSL)
endif()
```

And for lib/CMakeLists.txt, this way

```cmake
set (LIBRARIES "")

if (WITH_TLS)
    set (LIBRARIES OpenSSL::SSL)
endif()
```

besmarsh:
Recent CMake changes caused CMake builds with the `WITH_TLS` option set to `OFF` to fail. The OpenSSL package is only found (`find_package()`) if `WITH_TLS` is `ON`, but linking to OpenSSL for the broker and library is not guarded by `WITH_TLS`. The build therefore fails.

Guard linking to OpenSSL, only linking if `WITH_TLS` is set.

Fixes #3318

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -60,7 +60,9 @@ set(C_SRC
 	util_mosq.c util_topic.c util_mosq.h
 	will_mosq.c will_mosq.h)
 
-set (LIBRARIES OpenSSL::SSL)
+if (WITH_TLS)
+	set (LIBRARIES OpenSSL::SSL)
+endif()
 
 if (UNIX AND NOT APPLE AND NOT ANDROID)
 	find_library(LIBRT rt)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -163,7 +163,10 @@ if (WITH_DLT)
     set (MOSQ_LIBS ${MOSQ_LIBS} ${DLT_LIBRARIES})
 endif (WITH_DLT)
 
-set (MOSQ_LIBS ${MOSQ_LIBS} OpenSSL::SSL)
+if (WITH_TLS)
+	set (MOSQ_LIBS ${MOSQ_LIBS} OpenSSL::SSL)
+endif()
+
 # Check for getaddrinfo_a
 include(CheckLibraryExists)
 check_library_exists(anl getaddrinfo_a  "" HAVE_GETADDRINFO_A)


GasparQ:
Could this changes be merged quickly ?

ralight:
Thank you

PR #3341, opened 2025-08-14
Fix expiry list corruption

nickajacks1:
The internal expiry list can be corrupted if a client connects with a nonzero session expiration time, they disconnect, a plugin kicks said client, and mosquitto expires their session (in that order). The issue arises because the client context is added to the expiry_list twice. Fix this by first checking if the client is already on the list before adding them again.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -235,7 +235,9 @@ void context__disconnect(struct mosquitto *context)
 				context__add_to_disused(context);
 			}
 		}else{
-			session_expiry__add(context);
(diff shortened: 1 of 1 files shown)


ralight:
Thank you very much for this. I'd prefer to make the check inside session_expiry__add() instead, so I've made that change and added a test for it, I hope you don't mind. The commits are 0f96d25b747bb607283688dfac2210f06f109c88 and f604b09f57d7fa77ecdeedc17e0c31adf54a0803

PR #3343, opened 2025-08-16
feat: update mosquitto strings

Its-Just-Nans:

-----

Hello

- Add the correct connack codes (add fixes the usage of it)
- Add a function for disconnect

### DIFF
--- a/apps/mosquitto_ctrl/client.c
+++ b/apps/mosquitto_ctrl/client.c
@@ -106,12 +106,8 @@ static void on_connect(struct mosquitto *mosq, void *obj, int reason_code, int f
 			ctrl->response_topic = NULL;
 		}
 	}else{
-		if(ctrl->cfg.protocol_version == MQTT_PROTOCOL_V5){
-			if(reason_code == MQTT_RC_UNSUPPORTED_PROTOCOL_VERSION){
-				fprintf(stderr, "Connection error: %s. Try connecting to an MQTT v5 broker, or use MQTT v3.x mode.\n", mosquitto_reason_string(reason_code));
-			}else{
-				fprintf(stderr, "Connection error: %s\n", mosquitto_reason_string(reason_code));
-			}
+		if(ctrl->cfg.protocol_version == MQTT_PROTOCOL_V5 && reason_code == MQTT_RC_UNSUPPORTED_PROTOCOL_VERSION){
+			fprintf(stderr, "Connection error: %s. Try connecting to an MQTT v5 broker, or use MQTT v3.x mode.\n", mosquitto_connack_string(reason_code));
 		}else{
 			fprintf(stderr, "Connection error: %s\n", mosquitto_connack_string(reason_code));
 		}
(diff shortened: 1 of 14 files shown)


ralight:
I don't think I agree with this change. Could you explain what problem you believe it to be solving?

Its-Just-Nans:
> I don't think I agree with this change. Could you explain what problem you believe it to be solving?

`mosquitto_connack_string()` is incorrect for the MQTT v5 as it currently handle only code from 0 to 5

https://github.com/eclipse-mosquitto/mosquitto/blob/77868330e14e272fc69a54aa1bc29534bfb538cb/libcommon/strings_common.c#L148-L166

Currentlyt, to get the string of a connack code, we should do
- `mosquitto_connack_string()` for MQTTv3
- `mosquitto_reason_string()` for MQTTv5

And it's also not documented if I'm correct - only as comment here https://github.com/eclipse-mosquitto/mosquitto/blob/77868330e14e272fc69a54aa1bc29534bfb538cb/examples/subscribe/basic-1.c#L17-L20


This fix, update the `mosquitto_connack_string` for both 
- MQTTv3 `3.2.2.3 Connect Return code` on https://docs.oasis-open.org/mqtt/mqtt/v3.1.1/os/mqtt-v3.1.1-os.html
- MQTTv5 `3.2.2.2 Connect Reason Code
` on https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901079

Also I've added strings for disconnect code translation using `mosquitto_disconnect_string()` for the `3.14.2.1 Disconnect Reason Code
` on https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901208
also adding test, and fixing examples

ralight:
Ah, thank you I understand. I believe the answer is to better document it then. There is no such thing as a connack code/string in MQTT v5.0, and no such thing as a reason code/string in MQTT v3.1.1, so mixing the two doesn't make sense to me.

Its-Just-Nans:
Hum that's strange because on MQTTv5 (https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html) I see
- `2.4 Reason Code`
- `3.2.2.2 Connect Reason Code`

On MQTTv3 I see
- `3.2.2.3 Connect Return code`

On MQTTv5, the codes from 2.4 have a general name e.g
| Paragraph                                         | code |
|:------------------------------------------------:|:---------------------------------------------------------:|
|`2.4 Reason Code` | `155 - 0x9B - QoS not supported` |
| `3.2.2.2 Connect Reason Code` | `155 - 0x9B - QoS not supported - The Server does not support the QoS set in Will QoS.` |
|`3.14.2.1 Disconnect Reason Code`  | `155 - 0x9B - QoS not supported - The Client specified a QoS greater than the QoS specified in a Maximum QoS in the CONNACK.` |

What do you think about the `mosquitto_disconnect_string()`?

What do you think about the `mosquitto_connack_string()` handling also MQTTv5 code?
For example `if(code > 5) return mosquitto_reason_string(code)`

I can also edit the documentation if needed

PR #3358, opened 2025-08-29
Add configure-time check for -lanl

titouanc:
Since glibc 2.34, libanl features have been integrated directly into libc [1].
For backward compatibility, some toolchains still provide a shim for
libanl as a separate .so, but new toolchains (for example for new archs
like loongarch) do not provide it anymore.

In such a case, building mosquitto fails at link time with (see [2])
    > cannot find -lanl: No such file or directory

To fix this problem while maintaining compatibility with older toolchains,
check if a simple program that uses libanl can be compiled without -lanl,
and only add the linker flag otherwise.

[1] https://sourceware.org/pipermail/libc-alpha/2021-August/129718.html
[2] https://autobuild.buildroot.org/results/16223cd838876abc9b6f941f7dc20d23afa32c3b/build-end.log

Signed-off-by: Titouan Christophe <titouan.christophe@mind.be>

-----

### DIFF
--- a/config.mk
+++ b/config.mk
@@ -318,8 +318,11 @@ ifeq ($(WITH_EC),yes)
 endif
 
 ifeq ($(WITH_ADNS),yes)
-	BROKER_LDADD:=$(BROKER_LDADD) -lanl
 	BROKER_CPPFLAGS:=$(BROKER_CPPFLAGS) -DWITH_ADNS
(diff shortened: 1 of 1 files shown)


ralight:
Thanks for this, I wasn't aware the requirement had gone away.

I've implemented it slightly differently and also added cmake support in 9be6f9a409585b579efa4002c7a5da26ae6a996a and dfde35aa0fa5863863f4d0c8f2bbefe39320a119 for the develop and fixes branches, so I'm going to close this. Thanks again!

PR #3387, opened 2025-10-09
Support delayed auth with extended authorization

### LINKED ISSUE #3357: Delayed Auth for Extended Auth
MOSQ_ERR_AUTH_DELAYED is only recognized for basic auth. However, it would be useful in extended auth contexts as well.
In my case, both basic and extended auth use the same authentication backend process, which can take some time (we prefer extended auth because it allows a client to reauthenticate periodically). As such, it is desirable to have this authentication be asynchronous, and returning MOSQ_ERR_AUTH_DELAYED is the most natural choice for this.

Sadly, for extended auth, we'll probably have to do a workaround where the server immediately sends a reply AUTH that effectively tells the client to check back after N milliseconds, repeating the back & forth until the process is complete. It would be nice to be able to simply return MOSQ_ERR_AUTH_DELAYED instead.

I will look into potentially contributing this change (time permitting), but I figured I would create an issue in case there was a particular reason extended auth does not support this option or in case I was missing some simple alternative.

PS: it would be somewhat awkward to call `mosquitto_complete_basic_auth` for an extended auth attempt...

nickajacks1:
Support MOSQ_ERR_AUTH_DELAYED for extended auth.
Added `mosquitto_complete_extended_auth`, which optionally sends auth data.  Sadly, there's a bit of code duplication between this and `handle__auth`. I wasn't sure if it was worth splitting out into a shared segment of code or not.
Added two client states, `mosq_cs_delayed_ext_auth` and `mosq_cs_delayed_ext_reauth`. This seemed to be cleaner than trying to use existing client states.
Updated the `delayed_auth` plugin example to support extended auth.

I commandeered `contexts_by_id_delayed_auth` a bit. My understanding is that previously, that table was disjoint with `contexts_by_id`, but that won't work well for reauthentication. Since the same hash handle can't be used for more than one table, I added a new handle.

Maintainers, feel free to modify/discard any portion of this changeset as necessary.

Fixes #3357


nickajacks1:
I will be available to resolve conflicts only after Nov 1 at the earliest

nickajacks1:
I've resolved the conflicts, but I still need to go back and double check my work. I don't trust the me from 3 months ago.

nickajacks1:
Closing because my org's priorities changed and I don't plan to revisit this in the near term. Maybe eventually.

PR #3426, opened 2025-12-11
bridge server side ws:// protocol support added for libwebsocket build

miroslav-mastny-regosys:

-----

### DIFF
--- a/src/mosquitto_broker_internal.h
+++ b/src/mosquitto_broker_internal.h
@@ -915,6 +915,9 @@ int mosquitto_acl_check(struct mosquitto *context, const char *topic, uint32_t p
 int mosquitto_basic_auth(struct mosquitto *context);
 int mosquitto_psk_key_get(struct mosquitto *context, const char *hint, const char *identity, char *key, int max_key_len);
 
+void  set_server_ext_data(SSL* ssl, struct mosquitto* context, struct mosquitto__listener* listener);
+unsigned int psk_server_callback(SSL* ssl, const char* identity, unsigned char* psk, unsigned int max_psk_len);
+
 int mosquitto_security_init_default(void);
 int mosquitto_security_apply_default(void);
 int mosquitto_security_cleanup_default(void);
--- a/src/net.c
+++ b/src/net.c
@@ -304,7 +304,14 @@ static int client_certificate_verify(int preverify_ok, X509_STORE_CTX *ctx)
 #ifdef FINAL_WITH_TLS_PSK
 
 
-static unsigned int psk_server_callback(SSL *ssl, const char *identity, unsigned char *psk, unsigned int max_psk_len)
+void  set_server_ext_data(SSL* ssl, struct mosquitto* context, struct mosquitto__listener* listener)
+{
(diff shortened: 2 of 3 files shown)


ralight:
Thanks for this PR. I have a couple of comments:

* From what I can see, the description doesn't match the code. Could you please check it? 
* I don't want to support libwebsockets in the future. There is a built-in websockets implementation that already supports the same TLS functionality as the normal listeners - perhaps this means the TLS-PSK changes aren't needed
* I'd want to have some tests to cover the new code

miroslav-mastny-regosys:
This patch add websocket protocol support to bridge server side. For client side is external tpc to websocket conversion tool still required.

This extension save server resources, there is not necessary websocket conversion back to tcp and allow transparent access throw one access point (web server) for applications, web pages and bridge. This solution is useable with preceding ssl termination server only (ngix or others). It removes complication for application with nonstandard ports (administrations and firewalls limitations) because https only protocol can be used.

We using this for some time, but new developed extension (build in websocket) is not compatible with "duplicit" ssl configuration (http server and bridge), this is reason why I push this feature now, before new buildin websocket will be developed finally. I don't know full future plans and supports libraries, but I did suppose then you can remove libwebsocket dependecy.

ralight:
Sorry, but maybe I'm being stupid but I can't match up the idea of adding websocket support to bridges with what you have here, most of which is related to setting TLS-PSK communication over an incoming websocket connection and is nothing to do with bridges.

Could you please provide some automated tests to demonstrate the use of the code? There are existing examples in `test/broker` but if you would like some hints please let me know.

miroslav-mastny-regosys:
So, there is bigger question "why do not add websocket support to bridge" :-) especially in world, where everething becomes  to by web.

You can imagine this. There is one mosquitto server with bridge interface and hundreds another installations which using mosquitto as local broker and connection to central server. In 99.9% organizations is no problem to access https, but specific bridge port to server need IT administrator support, answer the question "is that really necessary?", this is annoying and making troubles. Using websocket for bridge is elegant solution with one access to server. MQTT deliver messages, other applications access same server through https. Elegant and simple solution.

Of course there is double SSL processing (https and bridge PSK), this may look be extra and unnecessary, but let's leave it alone. New implementation of buildin websocket has one big stop. Currently ssl configuration using one shared configuration for websocket and for bridge separated by certificate configuration. That is why is pull limited to ws protocol and do not support wss. With buildin websocket is ssl configuration especially for websocket and there is missing structure or flag to use that configuration for bridge. It complicate bridge websocket implementation. So that's point of pull request. 

Finally would be great to support websocket bridge on listener and connections side, with ws and wss support for course. Without specific conversion tools. That's reason why there are no test available yet, connection side support is missing and external software for conversion tpc to websocket is required.

ralight:
I understand your arguments about the use of websockets. This PR does not do that though. All it does is allow incoming connections to tunnel TLS-PSK data over websockets. It has nothing to do with bridges.

There are no tests, and I'm sorry but I don't trust you based on your responses so I'm closing this PR.

miroslav-mastny-regosys:
Ok, accepter. You are wrong about bridge and TLS-PSK usage, so may be later again.

PR #3456, opened 2026-01-29
Add support for OpenSSL provider URIs

bdesplanq:
Mosquitto currently does not support OpenSSL 3 provider-based key URIs.
In OpenSSL 3 and later, URIs are used to reference private keys and key stores via the provider and STORE APIs (for example, pkcs11: URIs, but this can be any other URI scheme supported by a given provider). This mechanism replaces the legacy engine-based approach used in earlier OpenSSL versions.
According to the OpenSSL roadmap, engine support is deprecated and scheduled for removal in OpenSSL 4, requiring applications that rely on engines to migrate to provider-based implementations.
This pull request introduces a new keyform option, uri. When keyform is set to uri, Mosquitto uses the OpenSSL STORE API to load private keys via provider URIs, enabling compatibility with OpenSSL 3 providers and future OpenSSL releases.

The change has been validated  using either a URI private key reference at client side (e.g. mosquitto_sub) and at broker side.

----

-----

### DIFF
--- a/apps/mosquitto_ctrl/options.c
+++ b/apps/mosquitto_ctrl/options.c
@@ -623,7 +623,7 @@ int client_opts_set(struct mosquitto *mosq, struct mosq_config *cfg)
 	}
 #ifdef WITH_TLS
 	if(cfg->keyform && mosquitto_string_option(mosq, MOSQ_OPT_TLS_KEYFORM, cfg->keyform)){
-		fprintf(stderr, "Error: Problem setting key form, it must be one of 'pem' or 'engine'.\n");
+		fprintf(stderr, "Error: Problem setting key form, it must be one of 'pem', 'engine', or 'uri'.\n");
 		return 1;
 	}
 	if(cfg->cafile || cfg->capath){
--- a/client/client_shared.c
+++ b/client/client_shared.c
@@ -1370,6 +1370,11 @@ static int client_tls_opts_set(struct mosquitto *mosq, struct mosq_config *cfg)
 		SSL_CTX_set_keylog_callback(cfg->ssl_ctx, tls_keylog_callback);
 	}
 
+	/* keyform must be known before calling mosquitto_tls_set() */
+	if(cfg->keyform && mosquitto_string_option(mosq, MOSQ_OPT_TLS_KEYFORM, cfg->keyform)){
+		err_printf(cfg, "Error: Problem setting key form, it must be one of 'pem', 'engine' or 'uri'.\n");
+		return 1;
+	}
 	if(cfg->cafile || cfg->capath){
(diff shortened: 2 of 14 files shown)


bdesplanq:
I noticed a similar PR was just submitted that does the same:  #3488 . This illustrates there is a demand for provider support :-)
One difference I see is that I choose to add it with a new keyform value "uri" for backward compatibility and to keep the keyform parameter meaningful.
When keyform is pem, then there will still be an attempt to open the file to test that it exists in options.c
Once engines are fully removed from future OpenSSL versions, then the keyform parameter could disappear and uri can be the implied default...
@ralight If there is still an interest in this PR I can add a few test cases for code coverage. If you prefer to proceed with the more recent PR I can close this one instead...

amananas:
Ahhh yes, last time I checked your PR was not there yet, I should have checked again.

As you pointed out, I chose to avoid a new key type: store providers creation in openssl aims to remove any key type selection by the application, and only retrieve keys based on the URI or path and available providers, so adding a new type was akin to creating a future needed rework to me (at least that's my understanding, stop me there if I'm wrong). Here I wanted to directly follow openssl way of pulling the keys. It does remove the key existence check though, I'm not sure that's a big issue.

As you said, this could be done in a future version, and I'd gladly probe opinions on this, but I do have applications maintenance in mind: introducing a new 'uri' type will create deployment and maintenance tasks for people using it, and fully removing types later on as well.

bdesplanq:
@amananas  I'd be happy with either approach - quite similar in the end. My main interest is to have the provider store loading and uri support built-in in mosquitto since this is indeed the standard way of working for OpenSSL going forward. Then there is no need anymore to maintain this with patches...

ralight:
Hi both, just to let you know I'm not ignoring this but am letting the dust settle for a while from the 2.1 release before starting to look at new features for 2.2

cci-seidelt:
Hi @ralight , are there any news on including this feature? This would be a nice feature for us as well.

borisVanhoof:
@bdesplanq and @amananas, to help @ralight perhaps you can work towards 1 pull request that works for the both of you? I understand the benefit of https://github.com/eclipse-mosquitto/mosquitto/pull/3488 is that there is only the URI to configure.

Our use case:
To increase security on prpl Foundation's home gateway operating system (through OpenWRT) by using mosquitto with OpenSSL provider for pkcs#11. The secrets will be generated on and stored in a trusted environment (e.g. OP-TEE, because a hardware security module is more expensive). By setting pkcs#11 CKA_EXTRACTABLE, the private key cannot be read from the trusted environment. Exactly like https://github.com/eclipse-mosquitto/mosquitto/pull/719#issue-304460926 but with providers instead of engines (which are deprecated by now).

> Hi both, just to let you know I'm not ignoring this but am letting the dust settle for a while from the 2.1 release before starting to look at new features for 2.2

@ralight by when do you estimate you could look into this, so that we can organize ourselves accordingly?

PR #3513, opened 2026-02-18
Fix makefile build with WITH_SQLITE disabled

jesferman:
When using WITH_SQLITE=no, the build system correctly avoids to compile the plugin, but the install step fails because it tries to install it even if it has not been compiled.

This commit fixes this behaviour by totally omitting the plugin if WITH_SQLITE=no is specified, so no compilation and no installation is performed.

I followed the same build approach that is used with the option `WITH_PERSISTENCE` in `apps/Makefile`

I contributed the fix to `master` branch directly instead of `fixes` because the bug is not present there.

-----

### DIFF
--- a/plugins/Makefile
+++ b/plugins/Makefile
@@ -3,8 +3,11 @@ DIRS= \
 		dynamic-security \
 		examples \
 		password-file \
-		persist-sqlite \
 		sparkplug-aware
+
+ifeq ($(WITH_SQLITE),yes)
+DIRS+=persist-sqlite
+endif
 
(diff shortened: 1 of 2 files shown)


jesferman:
Should we also avoid to run the persistence tests when compiling with WITH_SQLITE=no?

ralight:
Thank you - I've closed it via https://github.com/eclipse-mosquitto/mosquitto/commit/c6669ef715cec7b884305cdf359a1c6ab1d447b9 and the following commit

PR #3526, opened 2026-02-25
Fix unecessary warning when inflight maximum is unlimited

fungiboletus:
Since Mosquitto 2.1, I get a lot of `Client %s has exceeded its receive-maximum quota. This behaviour must be fixed on the client.` warnings on the console, but I think those aren't necessary warnings as I run with the following configuration:

```
max_inflight_messages 0
max_queued_messages 0
```

In `context.c`, `inflight_quota` is set to `max_inflight_messages` which is 0 in my case.

I added a similar check like in `database.c`: `context->msgs_in.inflight_maximum != 0`.

I guess that should do it.

### DIFF
--- a/src/handle_publish.c
+++ b/src/handle_publish.c
@@ -107,7 +107,7 @@ int handle__accepted_publish(struct mosquitto *context, struct mosquitto__base_m
 	}
 
 	if(!cmsg_stored){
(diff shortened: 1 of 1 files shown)


404Wolf:
Hi! This has been a big issue for me, as every publish is generating this error log. What is the status of this?

PR #3568, opened 2026-04-15
Ca certificate reload

afreof:
tls: fix stale CA store not updated on SIGHUP certificate reload
    
listeners__reload_all_certificates() called only net__load_certificates()
on the existing SSL_CTX without calling net__tls_load_verify(). As a
result, any rotation of the CA bundle (cafile/capath) had no effect until
a full broker restart: clients whose certificate was signed by the new CA
were rejected, and clients whose certificate was signed by a revoked CA
 continued to be accepted.
    
Fix by recreating the SSL_CTX from scratch: net__tls_server_ctx() frees
and reallocates a fresh context, then net__tls_load_verify() loads CA
certs and CRLs into it. Loading into the existing context cannot work,
because its X509_STORE can only gain trust anchors and never drops the
ones that are no longer configured.
    
Listeners served by libwebsockets are skipped. Their SSL_CTX is owned by
libwebsockets and mosquitto only holds a borrowed pointer obtained from
SSL_get_SSL_CTX(), so freeing it here would release a reference that was
never taken. config__cleanup() already avoids this for the same reason.
    
Also free the client CA name stack on the cafile error path of
net__tls_load_verify(). That path is now reachable on every SIGHUP rather
than only at startup, so the leak would repeat on each reload.
    
Existing TLS connections are unaffected: SSL objects hold a reference
to their parent SSL_CTX, so the old context is kept alive until the
last connection derived from it is closed.

Tests added:
- add SIGHUP TLS chain-stability regression test (test coverage to lover risk in related code but not directly affected by this change, theoretically)
- add SIGHUP CRL reload regression test (test coverage to lover risk in related code but not directly affected by this change, theoretically)
- add CA-rotation on SIGHUP regression test (test for this change)

-----


azurit:
@ralight Any hints on when this can be merged? The fix looks pretty simple and straightforward.

We are experiencing similar issues: Certificate rotation requires a restart, which results into all connections being dropped. Need a softer way to handle this.

afreof:
Rebased on top of latest commit from master branch. While doing this, I tried to adapt the code to follow the new conventions for tests which got merged 2 days ago. I hope this simplifies the re-view and allows to merge it soon.

afreof:
Hi @ralight

I tried to improve this again and I also run the TLS tests on valgrind. Is there a chance to get this merged or to get a feedback why not?

Thank you.

afreof:
> @ralight Any hints on when this can be merged? The fix looks pretty simple and straightforward.
> 
> We are experiencing similar issues: Certificate rotation requires a restart, which results into all connections being dropped. Need a softer way to handle this.

@azurit Would you be able to test or contribute to a review?

PR #3644, opened 2026-05-08
signal-args.py: don't kill init system, leading to system reboot

LocutusOfBorg:
Sending a kill to pid 1 is making the pbuilder containers reboot. Please fix the test, spawn a process for it, but don't play with init system PIDs


ralight:
Sorry about that, I only ever run in unprivileged environments. I'm closing this because I've got a better fix that improves the test as well.

